PR #18, opened 2021-11-24
Remove deprecated 'type' field

MikeGoldsmith:
There is a confusing duplicate of the spankind field called 'type'. This is not needed and should be removed.

- Resolves https://github.com/honeycombio/telemetry-team/issues/70
- Resolves https://github.com/honeycombio/telemetry-team/issues/71

- Remove the duplicate 'type' field when generating HNY events


vreynolds:
Do we need to coordinate customer communication on this? (In case there's people that use the `type` field in boards and saved queries)

MikeGoldsmith:
Yeah, it is possible queries using the type field exist. I'll reach out to CS on how they'd like us to move forward.

vreynolds:
won't fix https://github.com/honeycombio/telemetry-team/issues/71

PR #23, opened 2021-11-30
Update translate trace request calls to return struct instead of params

### LINKED ISSUE #21: Extend translation funcs to return SampleRate
**Is your feature request related to a problem? Please describe.**
We have duplicated code that extracts a sample rate from a set of attributes `map[string]interface` as seen in [Refinery](https://github.com/honeycombio/refinery/blob/3eedc4895aa0fc36ba8edea277c4337edf9bbc3d/route/route.go#L630) and [Hound](https://github.com/honeycombio/hound/blob/a5797ac2554d4b72e0d84b446862b7e3c2dbd20f/cmd/shepherd/app/otlp_traces.go#L337). It would be good if the returned map from the Translation functions returned the sample rate too.

It may also be nice to expose a custom type, and return typed fields (attributes, timestamp and samplerate) instead of a ma[string]interface{} that requires recasting to get the values.

For example, we currently do
```golang
attrs := ev["data"].(map[string]interface{})
timestamp := ev["time"].(time.Time)
```

and it would be nice if the fields were types:
```golang
attrs := ev.Attributes // map[string]interface{}
timestamp := ev.Timestamp // time.Time
sampleRate := ev.SampleRate // int
```

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

### LINKED ISSUE #22: Return request size when translating trace requets (gRPC & HTTP)
**Is your feature request related to a problem? Please describe.**
It would be useful to return the trace request size when translating so the caller can record and track request sizes.

**Describe the solution you'd like**
Update both `TranslateHttpTraceRequest` and `TranslateGrpcTraceRequest` to return the request size in addition to the batch of events

The request size can be retrieved using the following:
```go
size := int64(proto.Size(req))
```

**Describe alternatives you've considered**

**Additional context**

MikeGoldsmith:
Update both `TranslateHttpTraceRequest` & `TranslateGrpcTraceRequest` to return a known struct instead of a `map[string]interface{}`. Extend return struct to parse sampleRate attribute and calculate and return trace request size.

- Closes #21 
- Closes #22 

- Add `TranslateTraceRequestResult` and `Event` structs
- Return new structs as part of `TranslateHttpTraceRequest` & `TranslateGrpcTraceRequest` calls
- Extract sampleRate attribute and set on return struct
- Add tests for converting sample rate types, default values, etc
- Gets the request size using `proto.Size(request)` and sets on return struct
- Update tests to ensure the correct size is returned

### DIFF
--- a/otlp/trace.go
+++ b/otlp/trace.go
@@ -27,6 +27,7 @@ type TraceResult struct {
 type Event struct {
 	Attributes map[string]interface{}
 	Timestamp  time.Time
+	SampleRate int
 }
 
 func TranslateHttpTraceRequest(body io.ReadCloser, ri RequestInfo) (*TraceResult, error) {
@@ -85,6 +86,15 @@ func TranslateGrpcTraceRequest(request *collectorTrace.ExportTraceServiceRequest
 				}
 				if span.Attributes != nil {
 					addAttributesToMap(eventAttrs, span.Attributes)
+				}
+				var sampleRateKey string
+				if eventAttrs["sampleRate"] != nil {
+					sampleRateKey = "sampleRate"
+				} else if eventAttrs["SampleRate"] != nil {
+					sampleRateKey = "sampleRate"
+				}
+				if sampleRateKey != "" {
+					SampleRateValue := eventAttrs[sampleRateKey]
 				}
 
 				// copy resource attributes to event attributes
--- a/otlp/trace.go
+++ b/otlp/trace.go
@@ -19,18 +19,18 @@ const (
 	traceIDLongLength  = 16
 )
 
-func TranslateHttpTraceRequest(body io.ReadCloser, ri RequestInfo) ([]map[string]interface{}, error) {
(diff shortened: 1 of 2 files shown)


otlp/trace.go
```diff
@@ -17,21 +19,34 @@ import (
...
 	if err != nil {
 		return nil, ErrFailedParseBody
 	}
-	return TranslateGrpcTraceRequest(request)
+	return TranslateTraceRequest(request)
 }
 
-func TranslateGrpcTraceRequest(request *collectorTrace.ExportTraceServiceRequest) ([]map[string]interface{}, error) {
-	batch := []map[string]interface{}{}
+func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest) (*TranslateTraceRequestResult, error) {
```
vreynolds, line 48: This function doesn't have an error return case

vreynolds (approved):

otlp/trace.go
MikeGoldsmith: Yeah, I went with providing an error in case we need to do some error handling in the future and then we won't need to change the API.

PR #24, opened 2021-12-01
Add case insensitive tests for parsing headers

MikeGoldsmith:
Adds tests for checking case insensitivity when parsing headers into the `RequestInfo` struct.

- Adds case insensitivity tests


vreynolds (approved):

robbkidd (approved):
I want this, too.

PR #27, opened 2021-12-02
Downgrade otlp-proto to v0.9.0 & grpc to 1.40.0

MikeGoldsmith:
Downgrade otlp proto-go so we can use grpc v1.40.0

- downgrades otlp-proto-go to v0.9.0 and grpc to 1.40.0


robbkidd (approved):
go back

PR #35, opened 2022-02-02
feat!: environments & services support

vreynolds:

- support environment-aware OTLP instrumentation
- updates https://github.com/honeycombio/telemetry-team/issues/173

- breaking changes!
- ~~ValidateHeaders~~ >> ValidateTracesHeaders, ValidateMetricsHeaders
- TranslateTraceRequest will now return a map of dataset:[events], and will decide if the dataset should come form the metadata/header (legacy) or the service.name attribute

### DIFF
--- a/otlp/trace.go
+++ b/otlp/trace.go
@@ -25,7 +25,7 @@ const (
 
 type TranslateTraceRequestResult struct {
 	RequestSize int
-	Events      []Event
+	Batches     map[string][]Event
 }
 
 type Event struct {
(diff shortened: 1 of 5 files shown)


otlp/trace.go
```diff
@@ -35,25 +35,45 @@ type Event struct {
...
-	return TranslateTraceRequest(request)
+	return TranslateTraceRequest(request, ri)
 }
 
-func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest) (*TranslateTraceRequestResult, error) {
-	batch := []Event{}
+func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri RequestInfo) (*TranslateTraceRequestResult, error) {
+	if err := ri.ValidateTracesHeaders(); err != nil {
+		return nil, err
+	}
```
MikeGoldsmith, line 51: I didn't include header validation here before because our currently usage in refinery, Basenji & Shepherd wanted to validate headers before trying to translate the request.

Looking back, now I think it's fine to expect these funcs to perform the validation and consumers of Husky don't need to check themselves and instead expect validation errors to be returned from here.

MikeGoldsmith (approved):
Looks good 👍🏻

kentquirk:
This is tagged with "bump major" -- but we're still at 0.6. Is it time to call it 1.0? It might be, or we could still wait until E&S lands and the ground settles before making that decision.

MikeGoldsmith:
I'd like to see it being implemented in refinery, basenji & shepherd before calling it 1.0 just incase we need any modifications. When developing first time around, each implementation needed some changes so best to get it integrated then make it 1.0 👍🏻

PR #38, opened 2022-02-03
feat!: include individual batch sizes

vreynolds:

- shepherd needs to track how big each request is per dataset
- https://houndsh.slack.com/archives/C015UD6R8PJ/p1643848568491229

- approximating each dataset batch size to be `proto.Size(resourceSpan)`
- previously, `proto.Size(otlpRequest)` would be used in the same stat for the single dataset in the header

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -1,155 +1,155 @@
 package otlp
 
 import (
-    "context"
-    "encoding/json"
-    "net/http"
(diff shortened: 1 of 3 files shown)


MikeGoldsmith (approved):
Looks great 👍🏻

JamieDanielson (approved):

PR #40, opened 2022-02-03
prepare 0.8.0 release

vreynolds:
husky

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Husky Changelog
 
+## 0.8.0 2022-02-03
+
+## !!! Breaking Changes !!!
+
+- feat!: include individual batch sizes (#38)
+
 ## 0.7.0 2022-02-02
 
 ## !!! Breaking Changes !!!
--- a/version.go
+++ b/version.go
@@ -1,5 +1,5 @@
 package husky
 
 var (
-	Version string = "0.7.0"
+	Version string = "0.8.0"
 )


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Husky Changelog
 
+## 0.8.0 2022-02-03
+
+## !!! Breaking Changes !!!
+
+- feat!: include individual batch sizes (#38)
```
robbkidd, line 7: Is there any note we can add here to help the projects that use husky adjust to the breaking change?

kentquirk: That's Basenji, Refinery, and Shepherd, right? And our team is making the changes for all of them right now. Not that it would hurt to have it, but I'm honestly more eager to have the version stamped so I don't have to edit my go.mod file. :)

robbkidd: Yes. And OK.

kentquirk (approved):
We might consider revoking v0.7.0 just so no one ever accidentally depends on it.

robbkidd (approved):
judge me by my size, do you?

vreynolds:
> We might consider revoking v0.7.0 just so no one ever accidentally depends on it.

Isn't it kind of a no-go with go modules?

vreynolds:
There's nothing inherently broken with 0.7.0, as long as consumer doesn't need the batch size

kentquirk:
There is a command to revoke a revision in more recent versions of Go, but I guess it doesn't really matter.

PR #42, opened 2022-02-07
empty or missing service name should default to unknown_service

### LINKED ISSUE #41: E&S: set default unknown_service if service.name is empty or missing
In the E&S world currently, husky will only throw a "missing service.name" error if a particular resource span is missing a resource entirely or if the resource does not have a service.name attribute. We should return the same error when resource and attribute exist, but the attribute value is an empty string.

**Update**: In both scenarios of missing service name attribute and empty service name attribute, set the service name to `unknown_service` instead of erroring and dropping the data.

JamieDanielson:

- closes #41 

- Most SDKs will add a default service name if none are provided (`unknown_service:java`, etc). But in the (hopefully rare) scenario that a service name is missing or is an empty string, we should set the missing service name to `unknown_service`.
- this attempts to use `service.name` string value found in resource attributes. Service name will default to `unknown_service` if:
  - missing service name key or value in resource attributes
  - service name value is an empty string
  - service name value is not a string

Confirmed when sending empty service name string using otel-cli, no longer get "unable to resolve dataset error". Data is sent and new service name created `unknown_service`.

### DIFF
--- a/otlp/trace_test.go
+++ b/otlp/trace_test.go
@@ -851,10 +851,18 @@ func TestMissingServiceNameResourceUsesDefault(t *testing.T) {
 	assert.Equal(t, "unknown_service", batch.Dataset)
 }
 
-func TestEmptyServiceNameAttributeUsesDefault(t *testing.T) {
-	req := &collectortrace.ExportTraceServiceRequest{
-		ResourceSpans: []*trace.ResourceSpans{{
-			Resource: &resource.Resource{
+func TestEmptyOrInvalidServiceNameAttributeUsesDefault(t *testing.T) {
+	testCases := []struct {
+		name     string
+		resource *resource.Resource
+	}{
+		{
+			name:     "nil resource",
+			resource: nil,
+		},
+		{
+			name: "empty string",
+			resource: &resource.Resource{
 				Attributes: []*common.KeyValue{{
 					Key: "service.name",
 					Value: &common.AnyValue{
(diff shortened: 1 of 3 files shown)


vreynolds (approved):
❤️ nice test additions!

otlp/trace_test.go
```diff
@@ -779,7 +779,7 @@ func TestGetSampleRateConversions(t *testing.T) {
 	}
 }
 
-func TestMissingServiceNameAttributeReturnsError(t *testing.T) {
+func TestMissingServiceNameAttributeUsesDefault(t *testing.T) {
```
MikeGoldsmith, line 782: All of these tests could be combined into a single test with different test cases.

For example
```
testCases := []interface{
	name		string
	resource		Resource
	serviceName	string
}{
	{
		name: "nil resource",
		resource: nil,
		serviceName: "unknown_service"
	},
	{
		name: "empty string",
		resource: &resource.Resource{
			Attributes: []*common.KeyValue{{
				Key: "service.name",
				Value: &common.AnyValue{
					Value: &common.AnyValue_StringValue{StringValue: ""},
				},
			}},
		},
		serviceName: "unknown_service"
	},
	// .. bool, int, etc
}

for _, tc := range testCases {
	t.run(tc.name, func (t *testing.T) {
		// setup the test using tc.resource
		req := &collectortrace.ExportTraceServiceRequest{
			ResourceSpans: []*trace.ResourceSpans{{
				Resource: tc.resource,
				InstrumentationLibrarySpans: []*trace.InstrumentationLibrarySpans{{
					Spans: []*trace.Span{{
						TraceId: test.RandomBytes(16),
						SpanId:  test.RandomBytes(8),
						Name:    "test_span_a",
					}},
				}},
			}},
		}
		// test with tc.serviceName
		assert.Equal(t, "unknown_service", batch.Dataset)
	})
}
``

MikeGoldsmith (approved):
Change looks good. Added a suggestion on how to reorganise the tests.

MikeGoldsmith (approved):
smiling-husky

PR #52, opened 2022-03-02
feat: trim dataset when deriving from service.name

vreynolds:

- misconfigured service.name, such as ` ` or `\tmy-service\t`, when used as the dataset name, will result in weird behavior in the product (URL dataset will be a slug, hard to tell there's whitespace in the UI)
- updates https://github.com/honeycombio/telemetry-team/issues/196

- trim service.name value prior to applying it as the dataset

### DIFF
--- a/otlp/trace.go
+++ b/otlp/trace.go
@@ -80,19 +80,13 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 		if isLegacy {
 			dataset = ri.Dataset
 		} else {
-			if resourceSpan.Resource == nil {
+			serviceName, ok := resourceAttrs["service.name"].(string)
+			if !ok ||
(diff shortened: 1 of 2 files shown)


otlp/trace.go
```diff
@@ -80,19 +80,13 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 		if isLegacy {
 			dataset = ri.Dataset
 		} else {
-			if resourceSpan.Resource == nil {
```
vreynolds, line 83: simplified the branching here: 
- `resourceAttrs` will just be empty if `resourceSpan.Resource` is nil, so no need to pre-check
- collapsed ifs and elses so there one default case and one main case

MikeGoldsmith (approved):

PR #55, opened 2022-03-18
default sample rate to 1 if omitted or if 0

asdvalenzuela:

Our documentation states that sample rate is optional and defaults to 1. 

Our code uses a go int value for sample rate, and believes that the default zero value and omitted sample rates are the same. 

Our instrumentation shows sample rate as either 0 or 1, confusing the people reading the instrumentation and making math on sample rates difficult. (AVG(app.sample_rate) is not accurate because of mixed 0s and 1s influencing the average but not changing the system's behavior.)

Otel documentation has a [different meaning for zero](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#adjusted-count), but that's not how our system works. That they're different leads to confusion.

To clarify the current behavior for people reading code and looking at instrumentation, let's be explicit that both an omitted sample rate and sample rate explicitly set to zero are treated the same. Let's clarify by setting it to 1 in both cases. 

Co-authored-by: Michael Simo <michaelsimo@honeycomb.io>

### DIFF
--- a/otlp/trace.go
+++ b/otlp/trace.go
@@ -347,6 +347,12 @@ func getSampleRate(attrs map[string]interface{}) int32 {
 			sampleRate = math.MaxInt32
 		}
 	}
+	// To make sampleRate consistent between Otel and Honeycomb, we coerce all 0 values to 1 here
+	// A value of 1 means the span was not sampled
+	// For full explanation, see https://app.asana.com/0/365940753298424/1201973146987622/f
+	if sampleRate == 0 {
+		sampleRate = 1
+	}
 	delete(attrs, sampleRateKey) // remove attr
 	return sampleRate
 }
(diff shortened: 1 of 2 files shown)


reulan (approved):
Comment checks out with the link to Asana explains why sample rate is overridden to 1 at this level of the request. (though haven't done as though validate via poodle like we did for the other)

otlp/trace.go
```diff
@@ -347,6 +347,12 @@ func getSampleRate(attrs map[string]interface{}) int32 {
 			sampleRate = math.MaxInt32
 		}
 	}
+	// To make sampleRate consistent between Otel and Honeycomb, we coerce all 0 values to 1 here
+	// A value of 1 means the span was not sampled
+	// For full explanation, see https://app.asana.com/0/365940753298424/1201973146987622/f
+	if sampleRate == 0 {
+		sampleRate = 1
```
JamieDanielson, line 354: This seems right... I notice we have a `defaultSampleRate` variable already declared as`int32(1)`. Any reason to not use that instead of hardcoding 1 here?

asdvalenzuela: I didn't notice that, thank you! I'll update to use that value 👍

JamieDanielson (approved):
thanks @asdvalenzuela !

PR #58, opened 2022-03-31
Create dependabot.yml

vreynolds:

- husky does not get automatic dependency updates

- dependabot!


JamieDanielson (approved):
let the machine do it

Issue #68, opened 2022-04-26
OTLP library.version appears on spans from tracers that do not have a version set


robbkidd:
**Versions**

0.10.3

**Steps to reproduce**

1. Send a OTLP spans in batches.
2. Use a variety of span producing instrumentations (auto-instrumentation)
3. Use a custom tracer with a name set, but without a version set.
4. Witness that some spans from the version-less tracer will have a library.version attribute set on them. The version will match a version number from one of the other "libraries" (instrumentations) producing spans.

**Additional context**

[Reported](https://honeycombpollinators.slack.com/archives/CNQ943Q75/p1650904240619649) in Pollinators.

Using OpenTelemetry JS, a user noticed spans appearing Honeycomb with a `library.version` set on spans produced by their custom tracer on which no version was set.

My current theory is that [our guard](https://github.com/honeycombio/husky/blob/v0.10.3/otlp/trace.go#L99-L101) for setting `library.version` only if it is not an empty string (and `library.name` also if it weren't set) results in library spans without a version will reuse the version put into `resourceAttrs[]` by a previous iteration of the loop for a different library's spans earlier in the batch.

**Potential Solutions**

* Set `library.name` and `library.version` in `resourceAttrs[]` regardless of string length. The default value `""` is fine.
* Set `library.name` and `library.version` in `eventAttrs[]` [for each span](https://github.com/honeycombio/husky/blob/v0.10.3/otlp/trace.go#L109-L128) instead of setting them in `resourceAttrs[]` because `library.*` are _not_ resource attributes.

robbkidd:
Addressed in #69

PR #69, opened 2022-04-26
Fix library.version spillover between spans in a single batch

robbkidd:

- #68 

- add a test for spans from multiple libraries in a single batch
- change from setting `library.(name|version)` via `resourceAttrs` to setting on each event

### DIFF
--- a/otlp/trace.go
+++ b/otlp/trace.go
@@ -92,14 +92,6 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 
 		for _, librarySpan := range resourceSpan.InstrumentationLibrarySpans {
 			library := librarySpan.InstrumentationLibrary
-			if library != nil {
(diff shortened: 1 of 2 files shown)


robbkidd:
Test failure to demonstrate the issue.

robbkidd:
Ready for review. The test added in [41d2276](https://github.com/honeycombio/husky/pull/69/commits/41d22769ff450409e05511cd16f9859c7ebeeee5) and the CI failure for it demonstrates the problem described in the issue.

The fix is only one option.

Instead, we could continue to set `library.name` and `.version` in `resourceAttrs[]` if we delete them or set them to blank/nil at the start of each librarySpans loop. At the moment, I don't like this option very much because these library.* attributes are _not_ resource attributes and should not accidentally get set on all spans in a resourceSpans[].

MikeGoldsmith (approved):
Looks good 👍🏻

PR #74, opened 2022-05-04
Copy span attrs after resource attrs

### LINKED ISSUE #73: Which level of attributes should win when merging: Resource or Span?
On the occasion that a Span has an attribute name that matches a Resource attribute, [the current implementation has the Resource attribute's value winning](https://github.com/honeycombio/husky/blob/v0.10.4/otlp/trace.go#L129-L136) when attributes are gathered onto an event. Should that behavior remain? Or should we update the transform to assign Span attributes after Resource attributes so that the more specific value set is the one that wins? Raising the question because this precedence order will grow in layers if OTEP 201 is accepted and the Scope (neé InstrumentationLibrary) interstitial layer gets attributes, too.

MikeGoldsmith:
Ensures span attributes are preserved in favour of resource attributes that has the samekey. The current behaviour has resource attrs preserved.

- Closes #73

- Copy span attributes after resource attrs, they go into the same map so attrs with the same key are overwritten

### DIFF
--- a/otlp/trace.go
+++ b/otlp/trace.go
@@ -127,13 +127,14 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 					}
 				}
 
-				if span.Attributes != nil {
-					addAttributesToMap(eventAttrs, span.Attributes)
-				}
-
 				// copy resource attributes to event attributes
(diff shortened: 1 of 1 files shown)


robbkidd (approved):
I'm a huge fan.

a very large fan

PR #78, opened 2022-05-19
Convert OTLP span status.code type to integer earlier in the transform

robbkidd:

**Sign/Symptom:** 

A customer reported that `status_code` values had stopped appearing abruptly in one of their datasets. They had changed the type of the `status_code` column in the dataset schema from string to integer. _This is a totally reasonable thing to do because an OpenTelemetry span status code is an integer._

**Origin:**

Prior to this change, Husky sets `status_code` to a `trace.Status_StatusCode` type from the trace.pb.go generated from OTLP proto definitions. For Reasons™, that type _would_ convert to a string during our data ingest, but would _not_ convert to an integer (despite the type's association with actual `int32` constants).

We'll now do the type conversion from OTLP proto enum shenanigans to an `int` here in husky's transform of an OTLP span into an event. The events we produce from OTLP spans have no knowledge of or interest in the OTLP types generated from enums in the proto definitions.

### DIFF
--- a/otlp/trace.go
+++ b/otlp/trace.go
@@ -98,7 +98,7 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 				spanID := hex.EncodeToString(span.SpanId)
 
 				spanKind := getSpanKind(span.Kind)
-				status_code := evaluateSpanStatus(span.Status)
+				status_code, isError := evaluateSpanStatus(span.Status)
 
 				eventAttrs := map[string]interface{}{
 					"trace.trace_id":  traceID,
@@ -114,7 +114,7 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 				if span.ParentSpanId != nil {
 					eventAttrs["trace.parent_id"] = hex.EncodeToString(span.ParentSpanId)
 				}
-				if status_code == int(trace.Status_STATUS_CODE_ERROR) {
+				if isError {
 					eventAttrs["error"] = true
 				}
 				if span.Status != nil && len(span.Status.Message) > 0 {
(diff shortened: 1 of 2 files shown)


otlp/trace.go
```diff
@@ -257,22 +259,42 @@ func shouldTrimTraceId(traceID []byte) bool {
...
+	isError := false
+	status_code := int(status.Code)
+
+	switch status.Code {
+	case trace.Status_STATUS_CODE_OK:
+		// OK, thumbs up. Let's do this.
+	case trace.Status_STATUS_CODE_ERROR:
+		isError = true
+	case trace.Status_STATUS_CODE_UNSET:
+		//lint:ignore SA1019 keep DepCode until we move to proto v0.12.0+
```
MikeGoldsmith, line 288: I don't think we need the first part of this if statement, status_code will already be `0` from line 280.

otlp/trace.go
```diff
@@ -257,22 +259,42 @@ func shouldTrimTraceId(traceID []byte) bool {
...
+// See: https://github.com/open-telemetry/opentelemetry-proto/blob/59c488bfb8fb6d0458ad6425758b70259ff4a2bd/opentelemetry/proto/trace/v1/trace.proto#L230
+func evaluateSpanStatus(status *trace.Status) (int, bool) {
 	if status == nil {
-		return trace.Status_STATUS_CODE_UNSET
+		return int(trace.Status_STATUS_CODE_UNSET), false
 	}
-	if status.Code == trace.Status_STATUS_CODE_UNSET {
+
+	isError := false
+	status_code := int(status.Code)
```
MikeGoldsmith, line 280: Go idiom is to use use camelCase over snake_case :)

```suggestion
	statusCode := int(status.Code)
```

otlp/trace.go
```diff
@@ -257,22 +259,42 @@ func shouldTrimTraceId(traceID []byte) bool {
...
+		// OK, thumbs up. Let's do this.
+	case trace.Status_STATUS_CODE_ERROR:
+		isError = true
+	case trace.Status_STATUS_CODE_UNSET:
+		//lint:ignore SA1019 keep DepCode until we move to proto v0.12.0+
 		if status.DeprecatedCode == trace.Status_DEPRECATED_STATUS_CODE_OK {
-			return trace.Status_STATUS_CODE_UNSET
+			status_code = int(trace.Status_STATUS_CODE_UNSET)
+		} else {
+			status_code = int(trace.Status_STATUS_CODE_ERROR)
```
MikeGoldsmith, line 292:
```suggestion
			statusCode = int(trace.Status_STATUS_CODE_UNSET)
		} else {
			statusCode = int(trace.Status_STATUS_CODE_ERROR)
```

otlp/trace.go
```diff
@@ -257,22 +259,42 @@ func shouldTrimTraceId(traceID []byte) bool {
...
+			isError = true
 		}
-		return trace.Status_STATUS_CODE_ERROR
 	}
-	return status.Code
+
+	return status_code, isError
```
MikeGoldsmith, line 297:
```suggestion
	return statusCode, isError
```

otlp/trace.go
```diff
@@ -98,21 +98,23 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 				spanID := hex.EncodeToString(span.SpanId)
 
 				spanKind := getSpanKind(span.Kind)
+				status_code, isError := evaluateSpanStatus(span.Status)
```
MikeGoldsmith, line 101:
```suggestion
				statusCode, isError := evaluateSpanStatus(span.Status)
```

otlp/trace.go
```diff
@@ -98,21 +98,23 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
...
+
 				eventAttrs := map[string]interface{}{
 					"trace.trace_id":  traceID,
 					"trace.span_id":   spanID,
 					"type":            spanKind,
 					"span.kind":       spanKind,
 					"name":            span.Name,
 					"duration_ms":     float64(span.EndTimeUnixNano-span.StartTimeUnixNano) / float64(time.Millisecond),
-					"status_code":     getSpanStatusCode(span.Status),
+					"status_code":     status_code,
```
MikeGoldsmith, line 110:
```suggestion
					"status_code":     statusCode,
```

MikeGoldsmith (approved):
Looks good, nice work @robbkidd 👍🏻 

I've made a small suggestion for ignoring the first part of an if statement and updating a variable name to be more inline with go idioms.

MikeGoldsmith:
Will this affect users who have existing schemas where status_code is a string?

otlp/trace.go
robbkidd: haha … My Ruby is showing again.

robbkidd:
> Will this affect users who have existing schemas where status_code is a string?

I will confirm today, but I don't think so.

otlp/trace.go
robbkidd: 🤔 Yup. I agree.

Only need the `status.DeprecatedCode != trace.Status_DEPRECATED_STATUS_CODE_OK` else case

MikeGoldsmith (approved):

PR #82, opened 2022-06-14
Add OTLP logs support

MikeGoldsmith:
## Which problem is this PR solving?
Adds OTLP logs translation, with support for both OTLP structs (gRPC) and reading from a reader (HTTP).

- Closes https://github.com/honeycombio/telemetry-team/issues/234

## Short description of the changes
- Extract common utility funcs from trace.go into common.go
- Add new logs.go that exposes TranslateLogsRequest & TranslateLogsRequestFromReader
- Add unit tests to log_test.go

PR #87, opened 2022-07-14
Update to OTLP proto v0.18

MikeGoldsmith:

Updates the OTLP proto spec to latest (currently v0.18). This includes moving past a version where previously supported deprecated trace status' were allowed. This PR updates the logic to only use the newer status codes.

- Closes https://github.com/honeycombio/telemetry-team/issues/240

- Update otlp-proto package to 0.18
- Remove deprecated status code logic when processing spans
- Update unit tests for deprecated span status codes


robbkidd:
Blocking merge and release until log support (#82) merges and is released.

pkanal:
> Blocking merge and release until log support (#82) merges and is released.
> 
> Merge is also blocked until [transform of OTLP Metrics moves into husky](https://github.com/honeycombio/telemetry-team/issues/290). OTLP proto update and Metrics transform will travel together.

After working through it, we'll actually have to merge this PR before metrics can be moved into Husky. The order of operations will be:
1. Upgrade OTLP proto to v0.18 in Husky and release Husky (this PR)
2. Upgrade OTLP proto to v0.18 in Shepherd and upgrade to latest Husky and drop support for deprecated fields (https://github.com/honeycombio/hound/pull/10705)
3. Add metrics logic to Husky in a separate PR and release a new version of Husky
4. Update Shepherd to use Husky for metrics logic

MikeGoldsmith:
Unlocking now Logs has been used in Shepherd for a little while and we're confident it won't just break.

otlp/traces.go
```diff
@@ -248,30 +248,24 @@ func shouldTrimTraceId(traceID []byte) bool {
...
 func evaluateSpanStatus(status *trace.Status) (int, bool) {
+	const okStatus = int(trace.Status_STATUS_CODE_OK)
 	const unsetStatus = int(trace.Status_STATUS_CODE_UNSET)
 	const errorStatus = int(trace.Status_STATUS_CODE_ERROR)
 
 	if status == nil {
 		return unsetStatus, false
 	}
 
-	isError := false
```
kentquirk, line 258: This entire function reduces to:

```go
func evaluateSpanStatus(status *trace.Status) (int, bool) {
	if status == nil {
		return int(trace.Status_STATUS_CODE_UNSET), false
	}

	return int(status.Code), false
}
```

And since it never returns true, should probably also have the extra bool parameter removed.

otlp/traces_test.go
```diff
@@ -1184,44 +1184,32 @@ func TestEvaluateSpanStatus(t *testing.T) {
...
-		//   then the receiver MUST interpret the overall status to be STATUS_CODE_ERROR.
 		{
-			desc: "returns error when code is UNSET and deprecated_code is not OK",
+			desc: "returns error when code is ERROR",
 			status: &trace.Status{
-				Code:           trace.Status_STATUS_CODE_UNSET,
-				DeprecatedCode: trace.Status_DEPRECATED_STATUS_CODE_ABORTED,
-				Message:        "Old not OK!",
+				Code:    trace.Status_STATUS_CODE_ERROR,
+				Message: "Old OK!",
```
kentquirk, line 1200: Message should be something like "Error!"

otlp/traces_test.go
```diff
@@ -1184,44 +1184,32 @@ func TestEvaluateSpanStatus(t *testing.T) {
...
-		//   then the receiver MUST interpret the overall status to be STATUS_CODE_UNSET.
 		{
-			desc: "returns unset when code is UNSET and deprecated_code is OK",
+			desc: "returns unset when code is UNSET",
 			status: &trace.Status{
-				Code:           trace.Status_STATUS_CODE_UNSET,
-				DeprecatedCode: trace.Status_DEPRECATED_STATUS_CODE_OK,
-				Message:        "Old OK!",
+				Code:    trace.Status_STATUS_CODE_UNSET,
+				Message: "Old OK!",
```
kentquirk, line 1191: Message should be something like "Unset!"

otlp/traces_test.go
```diff
@@ -1184,44 +1184,32 @@ func TestEvaluateSpanStatus(t *testing.T) {
...
-		//   ignored, the `code` field is the sole carrier of the status.
 		{
-			desc: "returns code when code is not UNSET and deprecated_code is anything",
+			desc: "returns ok when code is OK",
 			status: &trace.Status{
-				Code:           trace.Status_STATUS_CODE_ERROR,
-				DeprecatedCode: trace.Status_DEPRECATED_STATUS_CODE_OK,
-				Message:        "Old OK!",
+				Code:    trace.Status_STATUS_CODE_OK,
+				Message: "Old OK!",
```
kentquirk, line 1209: Not "Old" anymore.

kentquirk (commented):
Since the old cruft is gone from upstream, we should remove it from here as well.

otlp/traces.go
MikeGoldsmith: We still want to return the bool if the status code is ERROR, but I agree this func is a little more complicated than necessary where it makes local constants of the ok, error and unset statuses. I'll refactor to make simpler.

kentquirk (approved):
tenor-83316687

PR #95, opened 2022-07-22
Only set meta.annotation_type field when log is part of trace

MikeGoldsmith:
It's important that meta.annotation_type is only set when a log is part of a trace, otherwise data can look strange in datasets that only contain logs.

- Only set the meta.annotation_type field when the log has a trace ID associated to it
- Add unit test to verify annotation_type is set correctly

### DIFF
--- a/otlp/logs_test.go
+++ b/otlp/logs_test.go
@@ -314,3 +314,56 @@ func TestLogsRequestWithInvalidBodyReturnsError(t *testing.T) {
 	assert.Nil(t, result)
 	assert.Equal(t, ErrFailedParseBody, err)
 }
+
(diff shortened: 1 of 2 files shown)


otlp/logs_test.go
```diff
@@ -314,3 +314,56 @@ func TestLogsRequestWithInvalidBodyReturnsError(t *testing.T) {
 	assert.Nil(t, result)
 	assert.Equal(t, ErrFailedParseBody, err)
 }
+
+func TestLogsWithoutTraceIdDoesNotGetAnnotationType(t *testing.T) {
```
pkanal, line 318: ty for this test 🎉

pkanal (approved):

PR #103, opened 2022-08-02
Add OTLP metrics support

MikeGoldsmith:
This PR extracts metrics ingest logic from Shepherd into Husky. This is consistent with how we handle traces & logs, Husky is the central place for this logic to live.
- Closes https://github.com/honeycombio/collection-team/issues/11

- Extracts metrics logic from [Shepherd](https://github.com/honeycombio/hound/blob/main/cmd/shepherd/app/otlp_metrics.go)
- Extracts metrics tests from Shepherd

## Differences between metrics & traces/logs
There are a few differences between the way certain things are handled for metrics compared to traces and logs because we want to keep things consistent as we move the data from Shepherd to Husky.
- Default `dataset` naming: logs/traces default to the service name for the dataset and if there is no service name it defaults to `unknown_service`. Metrics reads the dataset from the request info and if none is provided, it defaults to `unknown_metrics`. The request info logic is only present from traces & logs with legacy API keys. This is something we could change but I left it to match Shepherd for now.
- We have a lot of logic to combine metrics together so when we're creating events and batches we're doing so outside the loop of `resourceMetrics`. This is different to logs and traces where events are aggregated at the scope level and batches are aggregated at the resource level. Metrics in Shepherd didn't deal with batches and while we're emitting batches from Husky, there would only ever be one batch.

## Out of scope
This PR is meant to be a first step to moving this logic out of Shepherd which means we'll be making tweaks to this code as we integrate it with Shepherd. There are a few things that are out of scope for this PR:
- Handling telemetry data that Shepherd needs
- Support for kinesis which Shepherd will also need

## Tests to make sure this works
- [x] Added unit tests that test the core logic, extracted from Husky
- [x] Hooked into [corresponding Shepherd branch](https://github.com/honeycombio/hound/tree/purvi.metrics-extraction-husky) that calls these metrics functions and made sure all unit tests (except those to do with Shepherd's own telemetry passed)
- [x] On the Shepherd branch, emitted metrics locally and saw them successfully go through to Shepherd / local Honeycomb.


otlp/metrics.go
```diff
@@ -0,0 +1,372 @@
...
+					break
+				}
+
+				switch {
+				case i < len(bounds):
+					highestBound = bounds[i]
+				default:
+					// this is an odd scenario that we are not entirely sure creates
+					// valid data. We can use this to flag when investigation is needed
+					// potentially asking the customer for permission to look at the data
```
kentquirk, line 214: Feels like you're missing a telemetry hook here?

pkanal: Yes, we are. I was thinking we would keep telemetry out of scope for this particular PR because we will likely need a way to emit a bunch of telemetry data for metrics. We don't seem to emit telemetry for traces & logs from Husky itself so we'll need a way to get telemetry data we care about for metrics out of Husky for shepherd to use.

kentquirk: I don't mind if it's not part of this PR, but I don't think it's acceptable to push a change to Shepherd that will cause us to lose significant amounts of telemetry.

Maybe we can talk in the morning about finding a way to allow shepherd to inject a tracer into Husky.

MikeGoldsmith:
Closing for now - the metrics translation logic has change over the past year since this PR was created. We'll revisit in the future.

PR #108, opened 2022-08-12
Prepare release 0.14.0

pkanal:
Prepares release 0.13.1

- Updated version to 0.13.1
- Added changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
 # Husky Changelog
 
-## 0.13.1 2022-08-12
+## 0.14.0 2022-08-16
 
 ### Enhancements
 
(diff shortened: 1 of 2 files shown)


MikeGoldsmith:
v0.19 of the OTLP proto is now available. Let's update to that before releasing Husky.

MikeGoldsmith (changes requested):
I think with both updating the OTLP proto definitions to an internal copy and adding scope level attributes from spec version v0.19, we should do a minor bump to v0.14.0 here.

MikeGoldsmith (approved):
In the interest of keeping us moving, I've made the update to bump version to 0.14 and update the changelog 👍🏻

MikeGoldsmith:
This is currently blocked waiting for #111 to merge.

PR #111, opened 2022-08-16
Add support for scope attributes when translating traces & metrics requests

MikeGoldsmith:
Adds support for recording scope level attributes when translating OTLP traces and metrics requests.

- Closes https://github.com/honeycombio/telemetry-team/issues/322
- Unblocks #108 

- Adds scope attributes to the resource attributes list when translating trace and metrics requests
- Add unit tests to verify the attributes are being recorded correctly.


pkanal (approved):
🎉 🚢

PR #112, opened 2022-08-17
Add support for translating otlp/json trace and log requests

MikeGoldsmith:
This PR adds support for translating OTLP/JSON trace and log requests. This is achieve by checking the contentType header and using `protojson.Unmarshal` to decode the request into a protobuf message.

- Update trace and log header validation to allow application/json
- Use content type header to switch between protobuf and json unmarshal when decoding HTTP body
- Extend trace tests to test both protobuf and json requests can be decoded with different encodings 
- Add log tests to verify log requests can be decided with different encoding

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -14,6 +14,7 @@ import (
 	common "github.com/honeycombio/husky/proto/otlp/common/v1"
 	"github.com/klauspost/compress/zstd"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/encoding/protojson"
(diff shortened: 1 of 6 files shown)


otlp/common.go
```diff
@@ -246,7 +251,14 @@ func parseOtlpRequestBody(body io.ReadCloser, contentEncoding string, request pr
 		return err
 	}
 
-	err = proto.Unmarshal(bytes, request)
+	switch contentType {
+	case "application/protobuf":
```
vreynolds, line 255: should this case also have `application/x-protobuf`?

PR #113, opened 2022-08-18
Don't duplicate scope attributes between scopes

MikeGoldsmith:
Scope attributes were being carried over between scope entries which can lead to incorrect attributes being added to events. This change introduced a new utility func that extracts resource attributes and scope attributes and returns a new map[string]interface{} which is run per

- add getDataset to remove duplicate code
- move defaultService const to common
- make IsLegacy func a RequestInfo member
- add getResourceAttributes and getScopeAttributes utility funcs
- make sure we always apply attributes in order - resource ->  scope -> span/log

### DIFF
--- a/otlp/traces_test.go
+++ b/otlp/traces_test.go
@@ -53,8 +53,8 @@ func TestTranslateLegacyGrpcTraceRequest(t *testing.T) {
 			},
 			ScopeSpans: []*trace.ScopeSpans{{
 				Scope: &common.InstrumentationScope{
-					Name:    "instr_scope_name",
-					Version: "instr_scope_version",
+					Name:    "library-name",
+					Version: "library-version",
 					Attributes: []*common.KeyValue{
 						{
 							Key: "scope_attr",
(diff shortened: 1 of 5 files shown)


otlp/traces_test.go
```diff
@@ -137,8 +137,8 @@ func TestTranslateLegacyGrpcTraceRequest(t *testing.T) {
 	assert.Equal(t, "resource_attr_val", ev.Attributes["resource_attr"])
 	assert.Equal(t, 1, ev.Attributes["span.num_links"])
 	assert.Equal(t, 1, ev.Attributes["span.num_events"])
-	assert.Equal(t, "instr_scope_name", ev.Attributes["instrumentation_scope.name"])
```
vreynolds, line 140: Is `instrumentation_scope` --> `library` a breaking change we need to worry about?

vreynolds (approved):
Sneaky bug! 🐞 Looks good, as long as we're OK with just dropping the old instrumentation attribute.

MikeGoldsmith:
> Sneaky bug! 🐞 Looks good, as long as we're OK with just dropping the old instrumentation attribute.

Yeah, it was only introduced as part of scope attributes which we haven't received yet. Library was the original name and we duplicated to instrumentation_scope so isn't needed.

Issue #115, opened 2022-08-24
Should a name be synthesized somehow?


cartermp:
OTLP logs don't have a name as a part of the data model, unlike span events.

However, let's say there's multiple logs from the same span, each having a different body and timestamp. How should the user be able to sort out which to look at? Consider logs interleaved with span events here:

![image](https://user-images.githubusercontent.com/6309070/186439530-61b36fb9-25e4-41fa-acec-c106406bc939.png)

This is a comparison of the same data sent as a log vs. a span event. The user noted that the span events were easier to navigate because they had a name, and that the lack of a name for logs felt jarring.

Perhaps the right solution is something else entirely, such as a different piece of UI in the trace view for honeycomb that somehow represents logs the best.

cartermp:
Definitely like the prototype change.

pkanal:
Closed by https://github.com/honeycombio/hound/pull/11409

PR #116, opened 2022-08-24
[discussion] Synthesize a name for logs based on timestamp

cartermp:
Possibly for https://github.com/honeycombio/husky/issues/115 - this is to aid discussion rather than implement stuff.


MikeGoldsmith:
If we want a name for logs, I think this is a good option. We can always update later 👍🏻

Issue #118, opened 2022-08-30
Fallback to dataset setting for logs


cartermp:
There is a use case for logs that doesn't necessarily involve a service. Logs can come from somewhere else, then be forwarded via a collector and turned into OTLP logs.

If you're using Classic Honeycomb today, this use case is actually handled well, because the dataset is required. But if you use Environments, the dataset where the logs are sent to is called `unknown_service` even if you have the dataset specified.

There's a broader question of if we should respect dataset in the event of a a non-unknown `service.name` for all signals. But at least for Logs, there's a clear use case IMO.

robbkidd:
Idea for routing logs:
* if service.name set in a log entry, that’s the destination dataset
* no service.name attribute? use the x-honeycomb-dataset header value
* no x-honeycomb-dataset? use “unknown_log_source”?
  * let’s look at [the log spec](https://opentelemetry.io/docs/reference/specification/logs/data-model/)!

PR #121, opened 2022-09-07
[feat] allow OTLP/http-json for metrics

robbkidd:

Enabling support for all the OTLP content types. The ingest for metrics is still handled elsewhere at the moment, but that process uses husky's header validation before proceeding, so allowing JSON needs to happen here first.

* Refactored how supported content types are checked. A list of supported types is declared once and public function is provided to check for the presence of a given type in the supported list.
  * A note here that all three signals have the same validation implementation which I suspect may remain true for husky's level of abstraction. We could consider consolidating them down to one function.
* Bonus: included a correction to the HTTP error we return when a content type is not supported: [415 Unsupported Media Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415) instead of [501 Not Implemented](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501). If we don't support a content type, that's not a server side error.
* Refactored ingestion tests into loop-driven test cases of Classic/E&S, content types, and content encodings. Removed some tests that were duplicates of others.

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -105,22 +105,35 @@ func (ri RequestInfo) hasLegacyKey() bool {
 	return legacyApiKeyPattern.MatchString(ri.ApiKey)
 }
 
-// ValidateTracesHeaders validates required headers/metadata for a trace OTLP request
+// ValidateTracesHeaders checks for required headers/metadata for all OTLP requests
+// and specific to Traces.
 func (ri *RequestInfo) ValidateTracesHeaders() error {
-	if len(ri.ApiKey) == 0 {
-		return ErrMissingAPIKeyHeader
-	}
-	if ri.hasLegacyKey() && len(ri.Dataset) == 0 {
-		return ErrMissingDatasetHeader
-	}
-	if !IsContentTypeSupported(ri.ContentType) {
-		return ErrInvalidContentType
-	}
-	return nil // no error, headers passed all the validations
+	// validations specific to traces
+	// - none for now
+	// validations common for all OTLP requests
+	return validateHeaders(ri)
 }
 
-// ValidateMetricsHeaders validates required headers/metadata for a metric OTLP request
+// ValidateMetricsHeaders checks for required headers/metadata for all OTLP requests
+// and specific to Metrics.
 func (ri *RequestInfo) ValidateMetricsHeaders() error {
+	// validations specific to metrics
+	// - none for now
+	// validations common for all OTLP requests
+	return validateHeaders(ri)
+}
+
+// ValidateLogsHeaders checks for required headers/metadata for all OTLP requests
+// and specific to Logs.
+func (ri *RequestInfo) ValidateLogsHeaders() error {
+	// validations specific to logs
+	// - none for now
+	// validations common for all OTLP requests
+	return validateHeaders(ri)
+}
+
+// Header validation that is common across OTLP signal requests.
+func validateHeaders(ri *RequestInfo) error {
 	if len(ri.ApiKey) == 0 {
 		return ErrMissingAPIKeyHeader
 	}
@@ -131,20 +144,6 @@ func (ri *RequestInfo) ValidateMetricsHeaders() error {
 		return ErrInvalidContentType
 	}
 	return nil // no error, headers passed all the validations
-}
-
-// ValidateLogsHeaders validates required headers/metadata for a logs OTLP request
-func (ri *RequestInfo) ValidateLogsHeaders() error {
-	if len(ri.ApiKey) == 0 {
-		return ErrMissingAPIKeyHeader
-	}
-	if ri.hasLegacyKey() && len(ri.Dataset) == 0 {
-		return ErrMissingDatasetHeader
-	}
-	if !IsContentTypeSupported(ri.ContentType) {
-		return ErrInvalidContentType
-	}
-	return nil
 }
 
 // GetRequestInfoFromGrpcMetadata parses relevant gRPC metadata from an incoming request context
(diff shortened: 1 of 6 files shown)


otlp/logs_test.go
```diff
@@ -530,7 +391,7 @@ func TestLogsRequestWithInvalidContentTypeReturnsError(t *testing.T) {
 }
 
 func TestLogsRequestWithInvalidBodyReturnsError(t *testing.T) {
-	bodyBytes := test.RandomBytes(10)
+	bodyBytes := []byte{0x00, 0x01, 0x02, 0x03, 0x04}
```
robbkidd, line 394: Funny story: 10 random bytes in one local test run [produced a valid request body](https://en.wikipedia.org/wiki/Infinite_monkey_theorem) for OTLP logs.

otlp/traces_test.go
```diff
@@ -1258,67 +926,3 @@ func TestEvaluateSpanStatus(t *testing.T) {
 		})
 	}
 }
-
-func TestJsonRequest(t *testing.T) {
```
robbkidd, line 1262: This test seems thoroughly covered by the `application/json` encoding cases in `TestTranslateHttpTraceRequest` also updated in this file in this PR.

otlp/traces_test.go
```diff
@@ -1035,20 +717,13 @@ func TestEmptyOrInvalidServiceNameAttributeUsesDefault(t *testing.T) {
...
-
-			body := io.NopCloser(strings.NewReader(buf.String()))
 			ri := RequestInfo{
 				ApiKey:      "abc123DEF456ghi789jklm",
 				Dataset:     "legacy-dataset",
 				ContentType: "application/protobuf",
 			}
 
-			result, err := TranslateTraceRequestFromReader(body, ri)
+			result, err := TranslateTraceRequest(req, ri)
```
robbkidd, line 726: Because this tests is less about decoding and unmarshaling the request (well-covered in the HTTP test cases) and more about what to do given request headers and the state of the telemetry, I opted to snip out the marshal/unmarshal and go direct to our function for translating an OTLP proto struct.

A few tests were trimmed down along these lines.

otlp/logs_test.go
MikeGoldsmith: Yep, I've seen that before too 😬

otlp/common.go
```diff
@@ -83,10 +113,9 @@ func (ri *RequestInfo) ValidateTracesHeaders() error {
 	if ri.hasLegacyKey() && len(ri.Dataset) == 0 {
 		return ErrMissingDatasetHeader
 	}
-	switch ri.ContentType {
-	case "application/protobuf", "application/x-protobuf", "application/json":
+	if IsContentTypeSupported(ri.ContentType) {
 		return nil
-	default:
+	} else {
```
MikeGoldsmith, line 118: nit: I prefer not to end a func with an else statement and maybe even have `return nil` the last statement to help it look cleaner.

eg I would end these functions with
```golang
if !IsContentTypeSupported(ri.ContentType) {
    return ErrInvaludContentType
}
return nil
```

MikeGoldsmith (approved):
thumbs-up

otlp/common.go
robbkidd: Is it too early to unify the headers validation logic? If not and we can do it now, I'd say keep the separate signal-specific named validation functions as the public API surface in case the logic diverges again in the future.

```go
// ValidateTracesHeaders validates required headers/metadata for a trace OTLP request
func (ri *RequestInfo) ValidateTracesHeaders() error {
	return validateHeaders(ri)
}

// ValidateMetricsHeaders validates required headers/metadata for a metric OTLP request
func (ri *RequestInfo) ValidateMetricsHeaders() error {
	return validateHeaders(ri)
}

// ValidateLogsHeaders validates required headers/metadata for a logs OTLP request
func (ri *RequestInfo) ValidateLogsHeaders() error {
	return validateHeaders(ri)
}

func validateHeaders(ri *RequestInfo) error {
	if len(ri.ApiKey) == 0 {
		return ErrMissingAPIKeyHeader
	}
	if ri.hasLegacyKey() && len(ri.Dataset) == 0 {
		return ErrMissingDatasetHeader
	}
	if !IsContentTypeSupported(ri.ContentType) {
		return ErrInvalidContentType
	}
	return nil // headers passed all the wickets!
}
```

robbkidd: 👍🏻 [Reorganized](https://github.com/honeycombio/husky/pull/121/commits/fdf0a4143a330411ec3fb90a5a8354f141a3a973) to a series of if-guards with the else case as the final statement for each validation function.

❓ Then I bonus'd on a [consolidation](https://github.com/honeycombio/husky/pull/121/commits/d59fcd3a578b80305f4c15f418b0449ffcfe29cb) to try it on. I like it, but it can be easily snipped from this PR if folks don't want to dry that up.

MikeGoldsmith (approved):

otlp/common.go
JamieDanielson: I usually love the idea of consolidating instead of duplicating logic. However, I have concerns around shared validateHeader logic.

Metrics require dataset for both classic and current keys, at least for the foreseeable future (whereas traces require dataset for classic, service name for current).

Logs may have some [differing opinions](https://github.com/honeycombio/husky/issues/118) on whether to require service name or dataset (when not associated with a trace). I'm afraid we'd consolidate now just to split them back up again later.

robbkidd: Fair enough! I'll open another PR undoing that consolidation.

PR #122, opened 2022-09-09
[fix] unconsolidate header validation

robbkidd:

Our request header validation rules differ between OTLP Logs, Metrics, and Traces. We'll need to undo the consolidation of the validation logic done in #121. 🤭 

- [x] update tests to demonstrate the problem:
  * Metrics didn't cover "E&S key, no dataset" case.
  * There didn't appear to be tests for validating log requests.
- [x] make those tests pass!

### DIFF
--- a/otlp/common_test.go
+++ b/otlp/common_test.go
@@ -125,29 +125,38 @@ func TestAddAttributesToMap(t *testing.T) {
 
 func TestValidateTracesHeaders(t *testing.T) {
 	testCases := []struct {
+		name        string
 		apikey      string
 		dataset     string
 		contentType string
 		err         error
 	}{
-		{apikey: "", dataset: "", contentType: "", err: ErrMissingAPIKeyHeader},
-		{apikey: "a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1", dataset: "", contentType: "", err: ErrMissingDatasetHeader},
-		{apikey: "abc123DEF456ghi789jklm", dataset: "", contentType: "application/protobuf", err: nil},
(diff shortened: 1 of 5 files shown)


otlp/common_test.go
```diff
@@ -159,25 +168,68 @@ func TestValidateMetricsHeaders(t *testing.T) {
...
+		dataset     string
+		contentType string
+		err         error
+	}{
+		{name: "no key, no dataset", apikey: "", dataset: "", contentType: "", err: ErrMissingAPIKeyHeader},
+		{name: "no key, dataset present", apikey: "", dataset: "dataset", contentType: "", err: ErrMissingAPIKeyHeader},
+		{name: "classic/no dataset", apikey: "a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1", dataset: "", contentType: "", err: ErrMissingDatasetHeader},
+		{name: "classic/dataset present", apikey: "a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1", dataset: "dataset", contentType: "application/protobuf", err: nil},
+		{name: "E&S/no dataset", apikey: "abc123DEF456ghi789jklm", dataset: "", contentType: "application/protobuf", err: ErrMissingDatasetHeader},
+		{name: "E&S/dataset present", apikey: "abc123DEF456ghi789jklm", dataset: "dataset", contentType: "application/protobuf", err: nil},
```
robbkidd, line 213: For logs, I believe our current thinking is "set dataset in headers or service.name in the data or we have a fallback value" so these "yes key, no dataset" cases for logs can expect no error _for header validation_.

pkanal (approved):

cartermp (approved):

PR #124, opened 2022-09-12
For logs, fall back to dataset if service is unknown

cartermp:
fixes https://github.com/honeycombio/husky/issues/118

I was thinking about refactoring the tests so there wasn't so much duplication, but ... eh. 

So the ultimately fallback - `unknown_log_source` - won't actually get hit in practice because we ultimately require an `x-honeycomb-dataset` header. So I'm validating that in the test instead.

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -236,22 +236,16 @@ func getDataset(ri RequestInfo, attrs map[string]interface{}) string {
 
 func getLogsDataset(ri RequestInfo, attrs map[string]interface{}) string {
 	var dataset string
-	if ri.hasLegacyKey() {
-		dataset = ri.Dataset
-	} else {
-		serviceName, ok := attrs["service.name"].(string)
-		if !ok ||
-			strings.TrimSpace(serviceName) == "" ||
-			strings.HasPrefix(serviceName, "unknown_service") {
-			if strings.TrimSpace(ri.Dataset) == "" {
-				// In practice, this won't actually get hit, because we error on missing dataset header
-				dataset = unknownLogSource
-			} else {
-				dataset = ri.Dataset
-			}
+	serviceName, ok := attrs["service.name"].(string)
+	if !ok || strings.TrimSpace(serviceName) == "" || strings.HasPrefix(serviceName, "unknown_service") {
+		if strings.TrimSpace(ri.Dataset) == "" {
+			// In practice, this won't actually get hit, because we error on missing dataset header
+			dataset = unknownLogSource
 		} else {
-			dataset = strings.TrimSpace(serviceName)
+			dataset = ri.Dataset
 		}
+	} else {
+		dataset = strings.TrimSpace(serviceName)
 	}
 	return dataset
 }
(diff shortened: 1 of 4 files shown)


cartermp:
https://github.com/honeycombio/husky/pull/122 should be merged first and tests be updated

otlp/logs_test.go
```diff
@@ -25,52 +25,7 @@ func TestTranslateLogsRequest(t *testing.T) {
...
-				}, {
-					Key: "service.name",
-					Value: &common.AnyValue{
-						Value: &common.AnyValue_StringValue{StringValue: testServiceName},
-					},
-				}},
-			},
-			ScopeLogs: []*logs.ScopeLogs{{
-				Scope: &common.InstrumentationScope{
-					Name:    "library-name",
```
cartermp, line 45: Hmmm. In the refactor, this is changed to be `instr_scope_name` and `instr_scope_version`. But this test case is inconsistent with the one that follows it. Any ideas?

otlp/common.go
```diff
@@ -233,6 +234,22 @@ func getDataset(ri RequestInfo, attrs map[string]interface{}) string {
 	return dataset
 }
 
+func getLogsDataset(ri RequestInfo, attrs map[string]interface{}) string {
+	var dataset string
+	serviceName, ok := attrs["service.name"].(string)
+	if !ok || strings.TrimSpace(serviceName) == "" || strings.HasPrefix(serviceName, "unknown_service") {
+		if strings.TrimSpace(ri.Dataset) == "" {
+			// In practice, this won't actually get hit, because we error on missing dataset header
```
pkanal, line 242: I think we're switching to not erroring if there is no dataset, so this might get hit? Or do I have things backwards? 🙈

otlp/common_testhelpers.go
```diff
@@ -84,3 +90,54 @@ func testCaseNameForEncoding(encoding string) string {
 		return encoding
 	}
 }
+
+func buildExportLogsServiceRequest(traceID []byte, spanID []byte, startTimestamp time.Time, testServiceName string) *collectorlogs.ExportLogsServiceRequest {
```
pkanal, line 94: YESSSSSS 🙌🏽

pkanal (approved):
🚀

otlp/common.go
cartermp: Nah, you got it right. Should be resolved!

otlp/logs_test.go
robbkidd: Yup. We talked about it in a Zoom: this string is set only for the expectation for the value of `library.name` attributes on the resulting event. It can be whatever the test should expect, value is arbitrary so long as what goes in is what comes out.

cartermp:
Okay @robbkidd I think I sufficiently deleted enough nonsense from the copypasta that it's _close enough_ to be refactored into a decent shape.

robbkidd:
Offering my modifications combined with these changes over on #126

pkanal:
@robbkidd & @cartermp closing this in favour of #126

PR #126, opened 2022-09-13
[feat] dataset targeting rules for OTel Logs

### LINKED ISSUE #118: Fallback to dataset setting for logs

robbkidd:

- Resolves #118 

- Based on #124, this gathers the 3 test functions added there into a single, table-driven test function `TestLogs_DetermineDestinationDataset` to make the test cases clearer.
  - This does not include the `validateCommonLogsThings` helper function because the table-driven test for this behavior has only one expectation about the dataset to which the logs will be sent given the inputs.
- Adds several more test cases.


pkanal (approved):
Nice! 🚢

cartermp (approved):

PR #127, opened 2022-09-13
[fix] only classic teams must provide a dataset header when sending metrics

robbkidd:
More fallout from the header validation [consolidation](https://github.com/honeycombio/husky/pull/121/commits/3cba35dc0d16d840ee26c00ed9a3e5782b78cc29)-and-[diaspora](https://github.com/honeycombio/husky/pull/122).

E&S environments are not *required* to set x-honeycomb-dataset header; we have a fallback for them. Classic teams must provide a target dataset.


robbkidd:
It's a bug fix, but for a bug that is currently unreleased, so this change alone does not require a version bump.

cartermp (approved):
This comports with my understanding of the problem. Proceed!

PR #133, opened 2022-09-30
Add span sample rate to links and events

### LINKED ISSUE #132: Span Events and Links dont get Sample Rate attached
Right now the sample rate for a [span](https://github.com/honeycombio/husky/blob/main/otlp/traces.go#L92-L96) is added. However that same sample rate is not added to its [events](https://github.com/honeycombio/husky/blob/main/otlp/traces.go#L119-L122) and [links](https://github.com/honeycombio/husky/blob/main/otlp/traces.go#L146-L149). 

This has the consequence in that if you run a query against a span_event aggregates are wrong. 

This can be verified in our test where we do [test for a span](https://github.com/honeycombio/husky/blob/a9e37e48ecc4a91d78b50c2aa70281d37cfb6586/otlp/traces_test.go#L148) to have a sample rate, but the [events](https://github.com/honeycombio/husky/blob/a9e37e48ecc4a91d78b50c2aa70281d37cfb6586/otlp/traces_test.go#L164) and [links](https://github.com/honeycombio/husky/blob/a9e37e48ecc4a91d78b50c2aa70281d37cfb6586/otlp/traces_test.go#L175) are asserted to have a 0 Sample rate.

MikeGoldsmith:
The sample rate for a trace should be added to all events. Currently we only add it to the span and not it's span links and events.

This PR adds the span's sample rate to it's span links and events during processing.

- Fixes #132 

- Get the sample rate for the span and add to span link and span event structs
- Update existing test to verify the sample rate it set correctly

### DIFF
--- a/otlp/traces.go
+++ b/otlp/traces.go
@@ -86,13 +86,16 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 					addAttributesToMap(eventAttrs, span.Attributes)
 				}
 
+				// get sample rate after resource and scope attributes have been added
+				sampleRate := getSampleRate(eventAttrs)
+
 				// Now we need to wrap the eventAttrs in an event so we can specify the timestamp
 				// which is the StartTime as a time.Time object
 				timestamp := time.Unix(0, int64(span.StartTimeUnixNano)).UTC()
 				events = append(events, Event{
 					Attributes: eventAttrs,
 					Timestamp:  timestamp,
-					SampleRate: getSampleRate(eventAttrs),
+					SampleRate: sampleRate,
 				})
 
 				for _, sevent := range span.Events {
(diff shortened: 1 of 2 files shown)


otlp/traces.go
```diff
@@ -146,6 +150,7 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 					events = append(events, Event{
 						Attributes: attrs,
 						Timestamp:  timestamp, // use timestamp from parent span
+						SampleRate: sampleRate,
```
TylerHelmuth, line 153: @robbkidd and I talked about this some yesterday.  Is it safe to assume that the events and links on the span being sample are representative?  Is this a meaningful/safe assumption?

robbkidd: Since that conversation, I'm currently thinking "yes".

In a world where traces have been told to sample only 1/N of themselves, everything that occurs within a sampled trace is representative of N others.

Accurate? No. But that's head-sampling for you.

kentquirk: We have talked about this, and we agree that this strategy is the best way to handle this right now. Sampling is a dark art, and you can always construct scenarios under which it will fail, so we feel that consistency here is the best answer in the general case.

kentquirk (approved):
It's the best solution for now.

TylerHelmuth (approved):

PR #145, opened 2022-11-03
Fix kvlist and byte arrays marshalling

kentquirk:

This is step 1 of fixing a problem where some logs being sent to honeycomb contained some very large nested maps being sent as "Body" elements. 

Logs allow byte arrays to be sent, but we were just converting them to nil, so this adds byte array support and converts them with JSON marshal, which sends them as B64-encoded strings.

OTel kvlist is a special type that OTel invented to allow a map to be indexed and traversed in-order. But the defined semantics are that of a map. Our previous marshalling had two problems:
- It was sending the kvlist as a stringized array of objects, each of which had a single k-v pair.  In other words, `[{"a":1},{"b":2}]` -- but this is unneeded, as it's guaranteed that keys are unique. So this PR converts that to a simple map: `{"a":1, "b":2}`.
- A recursive routine was rendering each level of maps and arrays as strings, so unmarshalling nested objects would not result in an equivalent data structure. This PR creates a two-level recursion, where the first level converts objects to a "safe-for-honeycomb" data structure, and only then marshals the result with a single call to JSON marshal.

A followup PR will limit the marshalling length to something Honeycomb will consume.


MikeGoldsmith (approved):

otlp/common_test.go
```diff
@@ -334,3 +328,66 @@ func TestGetRequestInfoFromHttpHeadersIsCaseInsensitive(t *testing.T) {
...
+					{Key: "foo", Value: &common.AnyValue{Value: &common.AnyValue_IntValue{IntValue: 123}}},
+					{Key: "bar", Value: &common.AnyValue{Value: &common.AnyValue_DoubleValue{DoubleValue: 45.6}}},
+					{Key: "mom", Value: &common.AnyValue{Value: &common.AnyValue_StringValue{StringValue: "hi mom"}}},
+				},
+			}}}, `{"foo":123,"bar":45.6,"mom":"hi mom"}`},
+		{"nested map as mixed-type string", &common.AnyValue{
+			Value: &common.AnyValue_KvlistValue{KvlistValue: &common.KeyValueList{
+				Values: []*common.KeyValue{
+					{Key: "foo", Value: &common.AnyValue{Value: &common.AnyValue_IntValue{IntValue: 123}}},
+					{Key: "bar", Value: &common.AnyValue{Value: &common.AnyValue_DoubleValue{DoubleValue: 45.6}}},
```
TylerHelmuth, line 361: Any value in throwing an `AnyValue_BytesValue` entry into the nested maps case?

TylerHelmuth (approved):

PR #146, opened 2022-11-04
OTLP size fix

### LINKED ISSUE #147: Should we use jsoniter instead of encoding/json?
From @ianwilkes in a recent chat:

> but also we should use `jsoniter`, not `json`

Orthogonally to #145 and #146 we may want to look into using https://github.com/json-iterator/go. It looks like it can be a very simple drop-in replacement.

`jsoniter` purports to be over 150% faster at encoding in their benchmark, and json encoding is on our hot path for ingest. Does that track, or is it just some number on a table?

We could also use shepherd profiles in pyroscope to get a better sense of the data.

kentquirk:

- Part 2 of #145 
- This imposes a field size limit on values inbound from OTLP by cutting them off at a size limit in the marshaling process. This will cause them to fail to unmarshal as JSON, but since they're being sent as strings anyway, it doesn't seem like a very big deal.
- The size limit is 64K bytes, which is also imposed by Shepherd, but there's no point in doing the work if we're going to trim it anyway.
- Closes #147 

- A new limitedWriter type wraps strings.Builder for efficiency, and will only write up to the size limit. After reaching it, it lies to the caller and pretends it writes, but it doesn't. 
- Because we had to change to a json.Encoder to use this writer, we also have to cope with the fact that json.Encoder terminates its results with a newline, so some tests had to be changed to accommodate it.
- I tried to use a different, supposedly faster, json encoder but it broke several tests so I immediately abandoned it.
- I added some telemetry to see when this is being invoked -- it adds the fields `meta.truncated_bytes` and `meta.truncated_field` to the span if a field is truncated. This makes it visible to customers. We might also want to consider adding something visible in Dogfood if this happens but that's not a Husky change.
- I also took the recommendation in #147 to use a faster JSON encoder since we're already using it in Shepherd.

Fixes #147.


MikeGoldsmith (approved):

otlp/common.go
```diff
@@ -248,9 +253,36 @@ func getLogsDataset(ri RequestInfo, attrs map[string]interface{}) string {
...
+	w   strings.Builder
+}
+
+func newLimitedWriter(n int) *limitedWriter {
+	return &limitedWriter{max: n}
+}
+
+func (l *limitedWriter) Write(b []byte) (int, error) {
+	n := len(b)
+	if n+l.w.Len() > l.max {
```
TylerHelmuth, line 270: Is the limit in retriever also inclusive?

kentquirk: Yes. From hound:
```go
	if stringBytesLen > MaxVarstringLength {
		stringBytes = stringBytes[:MaxVarstringLength]
		stringBytesLen = MaxVarstringLength
	}
```

MikeGoldsmith:
I think this looks fine, and am happy for it to merge.

However, we need to be clear the additional attributes that are being added when a field is truncated are added to the converted honeycomb event that is stored in the customer's dataset, it's not added to Shepherd's telemetry.

PR #167, opened 2023-01-20
feat: OTLP bidirectional span links

puckpuck:
This adds support for bidirectional span links.  Currently, a span link is a single direction only, and there is no context in Honeycomb that a span is being linked to. Since span links are just another event in Honeycomb, this will add an additional span link event swapping the trace fields and linked trace fields for the attributes.

Draft: Right now, this works based on the presence of a span link attribute, but that requires customers to add that attribute and does not work for situations where the span link is generated from auto-instrumented code.  Before I write the unit tests for this, I would like to discuss if we use the special attribute or create the additional event all the time. I vote for always creating the additional event since this makes rolling out the functionality of bidirectional links automatic for all customers.

### DIFF
--- a/otlp/traces.go
+++ b/otlp/traces.go
@@ -162,6 +162,30 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 						Timestamp:  timestamp, // use timestamp from parent span
 						SampleRate: sampleRate,
 					})
+
+					if getBidirectionalLinkValue(attrs) {
+						// create a bidirectional link event
+						bidirectionalLinkAttrs := make(map[string]interface{})
+						for k, v := range attrs {
+							// swap the trace fields with the linked trace fields
+							if k == "trace.trace_id" {
+								k = "trace.link.trace_id"
+							} else if k == "trace.parent_id" {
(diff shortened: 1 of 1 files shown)


kentquirk:
Hey, Pierre. 

I think that because this adds events (which are billable), if we're going to do this we probably need it to be user-controllable. As you've done it here, that's fine because the user decides if they want to set that attribute on a link (although if they're going that far, they can also just do this themselves). But as you note, for autoinstrumentation that's not an option.

Making it configurable gets a little weird because Husky is used by both Refinery and Shepherd; Refinery could be controlled by a config (although we don't currently do anything like that for Husky). Shepherd would require some affordance in the Honeycomb UI. 

FWIW, this has come up in OTel discussions as well; it's a recognized issue, although no one has a great solution yet (and it gets extra complicated when sampling is involved). There's a tedious discussion [in this issue](https://github.com/open-telemetry/opentelemetry-specification/issues/2918#issuecomment-1314333564) of span links in the context of sampling.

MikeGoldsmith:
Closing as this PR has gone stale. We can re-open if we feel it is still something we want to do.

PR #170, opened 2023-02-16
feat: Detect if scope is from instrumentation library

MikeGoldsmith:
It's useful to know whether a span was created from an instrumentation library (eg OTel maintained library) or not (manual tracing or custom meter in app). This enables analysis of who is using which libraries and which libraries are most popular.

Using the library name and matching against a list of known prefixes allows us to detect an instrumentation library.

This PR updates the scope translation of OTLP data into a generic event to to test the scope's library name against a list of known prefixes. If detection returns `true`, it adds a new field `telemetry.instrumentation_field` to the list of returned attributes.

- Add known prefixes for java, python, .net, ruby, go and js
- Update  `common/getScopeAttributes` to test the library name against the prefixes
- If an instrumentation library matches, add `telemetry.instrumenation_library=true` to returned attributes
- Add test to verify a trace request adds the expected field for an instrumentation library scope
- Add test to verify detection works as expected for both known and unknown libraries names

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -51,6 +51,15 @@ var (
 
 	// Use json-iterator for better performance
 	json = jsoniter.ConfigCompatibleWithStandardLibrary
+
+	// List of per-lang instrumentation library prefixes
+	instrumentationLibraryPrefixes = []string{
+		"io.opentelemetry.instrumentation",            // Java,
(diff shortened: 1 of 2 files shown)


cartermp (approved):
Sparks joy!

TylerHelmuth (approved):

PR #172, opened 2023-02-21
fix: correct Ruby library/scope name prefix

robbkidd:

- #170 left OTel Ruby auto-instrumentation libraries undetected

- The OTel Ruby auto-inst libraries [report their scope/library name as a fully qualified Ruby class/module namespace string](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-ruby-contrib%20library.name&type=code).


JamieDanielson (commented):
Thanks for the inspiration! I updated Java and .net, so just need to rebase this against Main and update the `traces_test` for the new prefix

TylerHelmuth (approved):

PR #174, opened 2023-02-21
fix: update dotnet library name prefix

JamieDanielson:

- Missed detection of .NET instrumentation

- Updates instrumentation prefix for .NET instrumentation library names to be `OpenTelemetry.Instrumentation` instead of `opentelemetry.instrumentation`.
- Updates test for known value of an instrumentation library

While the instrumentation prefix is the same for both python and .net, case sensitivity makes them differ. Another option here could be to lowercase them, but this matches the string the way it's written today.

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -55,7 +55,8 @@ var (
 	// List of per-lang instrumentation library prefixes
 	instrumentationLibraryPrefixes = []string{
 		"io.opentelemetry",                            // Java,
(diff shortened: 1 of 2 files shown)


cartermp (approved):
omg it's the capitalization

JamieDanielson:
**Note** this was based on the Java branch so that should get merged in first, then this should auto-update to main as base

PR #179, opened 2023-03-07
fix: Proposed fix for JSON ingestion issue

kentquirk:

OTLP/JSON data coming direct from browser sources is arriving with traceID and spanID encoded as hex, [per spec](https://opentelemetry.io/docs/reference/specification/protocol/otlp/#json-protobuf-encoding). 

But when this is fed through the OTLP protojson library that Husky uses, these values are being decoded [as if they were encoded in base64](https://go.dev/play/p/rhP_HnZ2o3n).

The result is that Husky is corrupting these IDs.

- This adjusts the `BytesToTraceID` function and adds `bytesToSpanID` to special-case the conversion (which we can detect by its length).

### DIFF
--- a/otlp/traces.go
+++ b/otlp/traces.go
@@ -1,6 +1,7 @@
 package otlp
 
 import (
+	"encoding/base64"
 	"encoding/hex"
 	"io"
 	"math"
@@ -13,9 +14,11 @@ import (
 )
 
 const (
-	traceIDShortLength = 8
-	traceIDLongLength  = 16
-	defaultSampleRate  = int32(1)
+	traceIDShortLength     = 8
+	traceIDLongLength      = 16
+	traceIDCorruptedLength = 24
+	spanIDCorruptedLength  = 12
+	defaultSampleRate      = int32(1)
 )
 
 // TranslateTraceRequestFromReader translates an OTLP/HTTP request into Honeycomb-friendly structure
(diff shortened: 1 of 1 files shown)


otlp/traces.go
```diff
@@ -13,9 +14,11 @@ import (
 )
 
 const (
-	traceIDShortLength = 8
-	traceIDLongLength  = 16
-	defaultSampleRate  = int32(1)
+	traceIDShortLength     = 8
+	traceIDLongLength      = 16
+	traceIDCorruptedLength = 24
+	spanIDCorruptedLength  = 12
```
MikeGoldsmith, line 20: Could we rename these? Trace & span IDs aren't corrupted, they just aren't base64 encoded like when receiving via OTLP/protobuf.

otlp/traces.go
```diff
@@ -215,13 +218,31 @@ func BytesToTraceID(traceID []byte) string {
 		}
 	case traceIDShortLength: // 8 bytes
 		encoded = make([]byte, 16)
+	case traceIDCorruptedLength: // 24 bytes
```
MikeGoldsmith, line 221: nit: I think we could refactor this so we don't mix an early return for a special case with ones that expect to fall through.

MikeGoldsmith (approved):
I think this makes sense as OTLP/JSON spec shows traceID & spanID will only be hex and not base64 encoded.

The changes make functional sense but think the bytesToTraceID and bytesToSpanID funcs could be refactored to be a little cleaner. I also think we should rename the new consts for hex only encoded trace and span IDs -- they aren't corrupted, just not base64 encoded.

otlp/traces.go
kentquirk: 👍

otlp/traces.go
kentquirk: The choice was between hoisting the hex conversion into the other cases vs early return. I chose to minimize lines of change, but I'm happy to do it the other way.

pkanal (approved):
Thanks for the fix!

fchikwekwe (approved):
This seems good to me!

PR #184, opened 2023-03-09
fix: ingest base64-encoded spanIDs from browsers

robbkidd:

- Ingesting base64-encoded spanIDs emitted from browsers
- follow-up to #182 

- Added a test to demonstrate the issue
- `:=` -> `=` to fix unintended variable scoping


kentquirk (approved):
Thanks for the assist!

PR #191, opened 2023-04-20
feat: copy exception details from span event to parent span

cartermp:
This came from a discussion and screen share with @puckpuck

So, today, as we know, exceptions get captured as span events with OTel. That's fine.

However, it's not so fine for us, namely:

1. Querying them is a little awkward since you're often looking/caring about spans, but you need to actually query the span events. This makes it harder to do bubble ups and see what error/exception info differs.
2. Digging into a trace with an error yields no data on the span; you need to click into the span events tab (which people often miss!) and then find the span event that has the error details (there can be several that have no errors) and _then_ you can see the actual exception detail

This solves both problems by copying the exception details onto the parent span that has those events.

### DIFF
--- a/otlp/traces.go
+++ b/otlp/traces.go
@@ -120,16 +120,19 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 					}
 					if isError {
 						attrs["error"] = true
+					}
 
-						// Copy over span event error attributes because:
-						// 1. They are common and high-value for error investigations
-						// 2. It sucks to have to look at span events in our trace UI today to hunt these down on an error span
-						// 3. This makes bubble up better because you can see these error details without having to query the span events
+					// For span events that are following the "exception" semantic convention,
+					// we're going to copy their attributes to the parent span because:
+					// 1. They are common and high-value for error investigations
+					// 2. It sucks to have to look at span events in our trace UI today to hunt these down on an error span
+					// 3. This makes bubble up better because you can see these error details without having to query the span events
+					// If there is more than one exception event, only the first one we encounter will be copied.
+					if sevent.Name == "exception" {
 						for _, seventAttr := range sevent.Attributes {
-							if seventAttr.Key == "exception.message" ||
-								seventAttr.Key == "exception.type" ||
-								seventAttr.Key == "exception.stacktrace" ||
-								seventAttr.Key == "exception.escaped" {
+							switch seventAttr.Key {
+							case "exception.message", "exception.type", "exception.stacktrace", "exception.escaped":
+								// don't overwrite if the value is already on the span
 								if _, present := eventAttrs[seventAttr.Key]; !present {
 									eventAttrs[seventAttr.Key] = seventAttr.Value
 								}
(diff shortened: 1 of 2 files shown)


otlp/traces.go
```diff
@@ -125,6 +120,21 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 					}
 					if isError {
 						attrs["error"] = true
+
+						// Copy over span event error attributes because:
+						// 1. They are common and high-value for error investigations
+						// 2. It sucks to have to look at span events in our trace UI today to hunt these down on an error span
+						// 3. This makes bubble up better because you can see these error details without having to query the span events
+						for _, seventAttr := range sevent.Attributes {
+							if seventAttr.Key == "exception.message" ||
```
TylerHelmuth, line 129: Should we flatten all span events into Events?

otlp/traces.go
```diff
@@ -125,6 +120,21 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
...
+								seventAttr.Key == "exception.type" ||
+								seventAttr.Key == "exception.stacktrace" ||
+								seventAttr.Key == "exception.excaped" {
+								if _, present := attrs[seventAttr.Key]; !present {
+									attrs[seventAttr.Key] = seventAttr.Value
```
TylerHelmuth, line 134: `attrs` is the span event's attributes.  To set the attributes for the Honeycomb event representing the span we need to use `eventAttrs` right?

otlp/traces_test.go
```diff
@@ -182,6 +166,22 @@ func TestTranslateGrpcTraceRequest(t *testing.T) {
 			assert.Equal(t, "link", ev.Attributes["meta.annotation_type"])
 			assert.Equal(t, "span_link_attr_val", ev.Attributes["span_link_attr"])
 			assert.Equal(t, "resource_attr_val", ev.Attributes["resource_attr"])
+
+			// span
+			ev = events[2]
```
TylerHelmuth, line 171: Can we update/add tests that verify the error span event attributes are added to the honeycomb event for the span?

otlp/traces.go
cartermp: whoopsie

otlp/traces.go
kentquirk: Please do this with a switch statement instead of the ||.

otlp/traces.go
```diff
@@ -125,6 +120,21 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 					}
 					if isError {
 						attrs["error"] = true
+
```
kentquirk, line 123: The [semantic conventions for exceptions](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/exceptions/) say that the `name` field must be set to `exception`. I feel like testing that should be part of this. We set `isError` but don't look at that value -- and arguably, maybe we shouldn't do this if it's not leveraging that convention.

kentquirk (commented):
(I had started a comment that was addressed by reading further -- but I can't delete it, so I've edited it.)

otlp/traces.go
kentquirk: I've added this condition -- we only do this now if the span event name is "exception" (note that it's no longer doing it on everything that's named as an "error").

otlp/traces_test.go
kentquirk: I've added a test for that.

kentquirk:
@TylerHelmuth @cartermp In the interest of getting a new Husky release out soon, I've made the changes Tyler and I requested.

cartermp (commented):
I can't approve my own PR, but @kentquirk the changes look good to me!

TylerHelmuth (approved):

PR #219, opened 2023-12-07
feat: Add public functions for handling OTLP HTTP responses

TylerHelmuth:

OTLP HTTP protocol is very specific with how a server should respond to a client. This PR adds new functions to help handle OTLP HTTP response requirements:
1. Must set the content type to match in request content type
2. A status code must be set
3. For Success, a the response body MUST be a `Export[Trace|Metrics|Logs]ServiceResponse` with a nil PartialSuccess.
4. For errors, the response body MUST be a [Status](https://pkg.go.dev/google.golang.org/genproto/googleapis/rpc/status#Status).

See https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlphttp for full requirements.

- Adds several new functions to help users respond to HTTP OTLP requests.
- Adds new unit tests for these functions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 unit-tests.xml
+
+.idea
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"fmt"
 	"io"
 	"math"
 	"net/http"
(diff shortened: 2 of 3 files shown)


JamieDanielson (approved):
without rules theres chaos

kentquirk (approved):
I like this. Especially the tests.

PR #224, opened 2023-12-08
fix: bug where we could error after writing a status code

TylerHelmuth:

- `WriteOtlpHttpResponse` did a no-no where it could return an error after writing a status code to the Writer.  This makes it really hard for users to know the state of their Writer once the call to `WriteOtlpHttpResponse` ends.

- Update function order to no longer error once we've written a status code.
- Updates the comments.


kentquirk (approved):
I *almost* caught this in the previous review. I was looking at example code and noticed that even the Go authors never looked at the error value from Write...but didn't think it through all the way.

Good catch!

PR #230, opened 2024-01-02
fix: properly parse parent id for logs

kentquirk:

- In https://github.com/honeycombio/husky/pull/179 we added some code to deal with the fact that OTLP/JSON  processing can't differentiate between arrays of bytes that are encoded as hex and those that are encoded as base64. When JS sends us browser-side telemetry, it encodes IDs as hex, but when we receive it it's decoded as base64. This PR also applies that to logs.

- Adjust log's parentID so that it properly parses
- Add a new test case
- Move some code to common since it now applies across signals
- Move some tests to common
- Clarify some comments

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"math"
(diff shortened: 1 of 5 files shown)


otlp/common.go
```diff
@@ -476,3 +478,62 @@ func parseOtlpRequestBody(body io.ReadCloser, contentType string, contentEncodin
...
+		encoded = make([]byte, base64.StdEncoding.EncodedLen(len(spanID)))
+		base64.StdEncoding.Encode(encoded, spanID)
+	default:
+		encoded = make([]byte, len(spanID)*2)
+		hex.Encode(encoded, spanID)
+	}
+	return string(encoded)
+}
+
+func shouldTrimTraceId(traceID []byte) bool {
```
fchikwekwe, line 532: Nice. I didn't even think about this case.

fchikwekwe (approved):
I think this looks good to me.

PR #231, opened 2024-01-03
fix: Handle case of endtime before starttime

### LINKED ISSUE #211: Husky calculates a nonsense `duration_ms` when end time < start time
**Versions**
all, probably

**Steps to reproduce**

1. send an event where EndTimeUnixNano is before StartTimeUnixNano
2. Notice how in Honeycomb, the duration_ms is impossibly large
CleanShot 2023-10-19 at 16 46 04

**Additional context**
This is because Husky is simply subtracting start time from end time, [here](https://github.com/honeycombio/husky/blob/8649f0bc71709fe3441e89cbe669ae32da4efcb0/otlp/traces.go#L65)

It would be better if duration_ms was set to `0`.   If it's not too difficult, it would also be really nice to include a diagnostic field or capability of some kind, so that we/customers can identify negative duration spans without actually making them negative.  (e.g. a meta field or a derived column in a la `INGEST_TIMESTAMP()`)

kentquirk:

- #211

- Check for negative value and use 0 instead
- Add `meta.negative_duration = true` if that happens

Closes #211

### DIFF
--- a/otlp/traces.go
+++ b/otlp/traces.go
@@ -69,7 +69,7 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 				duration := float64(span.EndTimeUnixNano-span.StartTimeUnixNano) / float64(time.Millisecond)
 				if duration < 0 {
 					duration = 0
-					eventAttrs["meta.negative_duration"] = true
+					eventAttrs["meta.invalid_duration"] = true
 				}
 				eventAttrs["duration_ms"] = duration
 
--- a/otlp/traces.go
+++ b/otlp/traces.go
@@ -61,12 +61,18 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
 					"type":             spanKind,
 					"span.kind":        spanKind,
 					"name":             span.Name,
-					"duration_ms":      float64(span.EndTimeUnixNano-span.StartTimeUnixNano) / float64(time.Millisecond),
 					"status_code":      statusCode,
 					"span.num_links":   len(span.Links),
 					"span.num_events":  len(span.Events),
 					"meta.signal_type": "trace",
 				}
+				duration := float64(span.EndTimeUnixNano-span.StartTimeUnixNano) / float64(time.Millisecond)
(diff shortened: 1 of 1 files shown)


otlp/traces.go
```diff
@@ -61,12 +61,18 @@ func TranslateTraceRequest(request *collectorTrace.ExportTraceServiceRequest, ri
...
-					"duration_ms":      float64(span.EndTimeUnixNano-span.StartTimeUnixNano) / float64(time.Millisecond),
 					"status_code":      statusCode,
 					"span.num_links":   len(span.Links),
 					"span.num_events":  len(span.Events),
 					"meta.signal_type": "trace",
 				}
+				duration := float64(span.EndTimeUnixNano-span.StartTimeUnixNano) / float64(time.Millisecond)
+				if duration < 0 {
+					duration = 0
+					eventAttrs["meta.negative_duration"] = true
```
cartermp, line 72: Verging on bikeshedding, but why not a negative value like -1? It's just a little strange (to me at least) to write an event as having a negative duration when the duration isn't negative.

kentquirk: Mainly because when you do that, Honeycomb interprets it as a Very Large positive number.

I did consider doing this:

`eventAttrs["meta.negative_duration"] = -duration`

But thought that had the potential to be even more confusing.

Another option would be a different name. 'meta.end_before_start=true`?

cartermp: Ohhhh, I see. Well, that's fun. Carry on, LGTM!

JamieDanielson: Carrying on with bikeshedding I suppose, but what about something like `meta.invalid_duration`?

JamieDanielson (approved):
Non-blocking suggestion on the meta attribute. Also thinking a small unit test would be nice but a glance at the tests makes it seems like it's a nontrivial task for a trivial change, and since we're already testing for appropriate timestamps elsewhere it's probably not necessary.

kentquirk:
@JamieDanielson yes, the work to add a new test was going to be large, and I didn't feel it was justified. And I *love* the "invalid_duration" suggestion. Thanks!

PR #240, opened 2024-02-16
feat: Allow flattening of OTLP key-value lists

### LINKED ISSUE #235: Unpack and flatten array and key-value structures when feasible
This is related to https://github.com/honeycombio/husky/issues/90 but not the same issue.

Consider the following JSON representation of an incoming OTLP log:

```json
 {
    "body": {
      "clientIp": "111.222.888",
      "loglevel": "INFO",
      "message": "This is a test log message for abc application",
      "service": "helloworld",
      "span_id": "9c1544cc4f7ff369",
      "testtag": "fluentbit",
      "timestamp": "2024-01-08 18:37:08.150",
      "trace_id": "7ada6c95a1bd243fa9013cab515173a9"
    },
    "instrumentation.name": "",
    "instrumentation.version": "",
    "observed_timestamp": 0,
    "severity_text": "",
    "span_id": "",
    "trace_id": ""
  }
```

This _should_ be turned into an event with a dozen or so fields:

```json
 {
    "clientIp": "111.222.888",
    "loglevel": "INFO",
    "message": "This is a test log message for abc application",
    "service": "helloworld",
    "span_id": "9c1544cc4f7ff369",
    "testtag": "fluentbit",
    "timestamp": "2024-01-08 18:37:08.150",
    "trace_id": "7ada6c95a1bd243fa9013cab515173a9"
    "instrumentation.name": "",
    "instrumentation.version": "",
    "observed_timestamp": 0,
    "severity_text": "",
  }
```
 It is perhaps an open question of what we de-dupe, or if we need to call each of those nested fields `body.span_id` or `body.message` instead of `span_id` and `message`, but you get the idea. Seems useful, right? Haha, you fool! You fell for the trap! We do no such thing. Instead, it comes out looking more like this:

```json
 {
    "body":"{\"clientIp\":\"111.222.888\",\"loglevel\":\"INFO\",\"message\":\"This is a test log message for abc application\",\"service\":\"helloworld\",\"span_id\":\"9c1544cc4f7ff369\",\"testtag\":\"fluentbit\",\"timestamp\":\"2024-01-0818:37:08.150\",\"trace_id\":\"7ada6c95a1bd243fa9013cab515173a9\"}",
    "instrumentation.name": "",
    "instrumentation.version": "",
    "observed_timestamp": 0,
    "severity_text": "",
    "span_id": "",
    "trace_id": ""
  }
```

That's because when we see the `body` object, we serialize its contents as a JSON string. And so, as a result, sending us structured logs with some nice fields in the body (i.e., like a normal person would do) creates an unusable mess and a sucky user experience.

This is also true for trace data -- any field that's a key-value pair or array gets serialized.

Changing this is...technically a breaking change. The best kind. But in practice, people really don't rely on these structures too much. As it turns out, if we create garbage, people don't typically query garbage. There are a dozen or so Derived Columns in all of production Honeycomb that operate on a string like this, and most appear to be using regex to pluck out useful values...which looks like a workaround for this poor behavior. That's...really not a whole lot of dependency on this behavior, and from what I can glean, it really seems like people are working around this stuff rather than enjoying having to pluck out useful data from a giant string blob. Additionally, for logs in particular, the impact is so low that I don't think a single customer would be impacted.

### LINKED ISSUE #239: Consider unpacking OTLP arrays
From a customer request:

> [...] ]does Honeycomb accepts array attributes on OpenTelemetry? Their documentation mentions that as a valid attribute type, but when I try to send an array of strings from our testing instrumentation library it seems like what I'm getting is a string representation of the array instead. I'm still investigating if this is happening on the Ruby SDK - the string below suggests this might be the case since the string value in there is exactly what Ruby would output if it generated the string representation of the array in question - but I would appreciate any insights there.

Like in https://github.com/honeycombio/husky/issues/235, when we see an OTLP array we serialize the whole thing rather than add each element as an attribute.

Now, unlike key:value structures, arrays don't have named keys. So we'd have to consider naming each element based on an index. That might mean:

```
myArray: ["bananas", "kitty cats", "toilets"]
```

Would translate into:

```
{
  ...
  "myArray.0":"bananas",
  "myArray.1":"kitty cats",
  "myArray.2":"toilets",
}
```

MikeGoldsmith:
When translating OTLP kv lists we currently just encode as a JSON string which isn't particularly useful. This PR extends the extraction behaviour to traverse kv lists to generate multiple event fields. The max depth is set to 5 levels.

For example, this kv attribute:
```
key: "data"
value: {
  "key1": "val1",
  "key2": true,
}
```

Previously would be stored like this:
```
data: "{\"key1\": \"val1\","key2":true}"
```

After the change, the same kv list would be stored like this:
```
data.key1: "val1"
data.key2: true
```

This change exposes the `AddAttributesToMap` func for consumers to use directly. This will be useful until metrics translation can be moved into this library (eg from shepherd).

- Closes #235 

- Update logic to extract OTLP data types to traverse kv lists instead of always encoding as JSON string, with a max depth of 5
- Expose `AddAttributesToMap` as a public function so consumers can use directly
- Update log translation to set the `body` field with a JSON string of the attribute to preserve backwards compatibility
- Removes unused & broken truncation event fields

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -39,7 +39,7 @@ const (
 	unknownLogSource         = "unknown_log_source"
 
 	// maxDepth is the maximum depth of a nested kvlist attribute that will be flattened.
-	// If the depth is exceeded, the attribute should be added as a JSON string.
+	// If the depth is exceeded, the attribute should be added as a JSON string instead.
 	maxDepth = 5
 )
 
@@ -261,7 +261,7 @@ func getValueFromMetadata(md metadata.MD, key string) string {
 // AddAttributesToMap adds attributes to a map, extracting the underlying attribute data type.
 // Supported types are string, bool, double, int, bytes, array, and kvlist.
 // kvlist attributes are flattened to a depth of (maxDepth), if the depth is exceeded, the attribute is added as a JSON string.
-// Bytes and array values are added as JSON strings.
+// Bytes and array values are always added as JSON strings.
 func AddAttributesToMap(attrs map[string]interface{}, attributes []*common.KeyValue) {
 	for _, attr := range attributes {
 		// ignore entries if the key is empty or value is nil
@@ -403,7 +403,7 @@ func getMarshallableValue(value *common.AnyValue) interface{} {
 // addAttributeToMap adds an attribute to a map, extracting the underlying attribute data type.
 // Supported types are string, bool, double, int, bytes, array, and kvlist.
 // kvlist attributes are flattened to a depth of (maxDepth), if the depth is exceeded, the attribute is added as a JSON string.
-// Bytes and array values are added as JSON strings.
+// Bytes and array values are always added as JSON strings.
 func addAttributeToMap(result map[string]interface{}, key string, value *common.AnyValue, depth int) {
 	switch value.Value.(type) {
 	case *common.AnyValue_StringValue:
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -37,6 +37,10 @@ const (
 	gRPCAcceptEncodingHeader = "grpc-accept-encoding"
 	defaultServiceName       = "unknown_service"
 	unknownLogSource         = "unknown_log_source"
+
+	// maxDepth is the maximum depth of a nested kvlist attribute that will be flattened.
+	// If the depth is exceeded, the attribute should be added as a JSON string.
+	maxDepth = 5
 )
 
 // fieldSizeMax is the maximum size of a field that will be accepted by honeycomb.
@@ -254,30 +258,24 @@ func getValueFromMetadata(md metadata.MD, key string) string {
 	return ""
 }
 
-func addAttributesToMap(attrs map[string]interface{}, attributes []*common.KeyValue) {
+// AddAttributesToMap adds attributes to a map, extracting the underlying attribute data type.
+// Supported types are string, bool, double, int, bytes, array, and kvlist.
+// kvlist attributes are flattened to a depth of (maxDepth), if the depth is exceeded, the attribute is added as a JSON string.
+// Bytes and array values are added as JSON strings.
+func AddAttributesToMap(attrs map[string]interface{}, attributes []*common.KeyValue) {
 	for _, attr := range attributes {
 		// ignore entries if the key is empty or value is nil
 		if attr.Key == "" || attr.Value == nil {
 			continue
 		}
-		if val, truncatedBytes := getValue(attr.Value); val != nil {
-			attrs[attr.Key] = val
-			if truncatedBytes != 0 {
-				// if we trim a field, add telemetry about it; because we trim at 64K and
-				// a whole span can't be more than 100K, this can't happen more than once
-				// for a single span. If we ever change those limits, this will need to
-				// become additive.
-				attrs["meta.truncated_bytes"] = val
-				attrs["meta.truncated_field"] = attr.Key
-			}
-		}
+		addAttributeToMap(attrs, attr.Key, attr.Value, 0)
 	}
 }
 
 func getResourceAttributes(resource *resource.Resource) map[string]interface{} {
 	attrs := map[string]interface{}{}
 	if resource != nil {
-		addAttributesToMap(attrs, resource.Attributes)
+		AddAttributesToMap(attrs, resource.Attributes)
 	}
 	return attrs
 }
@@ -294,7 +292,7 @@ func getScopeAttributes(scope *common.InstrumentationScope) map[string]interface
 		if scope.Version != "" {
 			attrs["library.version"] = scope.Version
 		}
-		addAttributesToMap(attrs, scope.Attributes)
+		AddAttributesToMap(attrs, scope.Attributes)
 	}
 	return attrs
 }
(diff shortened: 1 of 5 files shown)


otlp/common.go
```diff
@@ -260,16 +263,13 @@ func addAttributesToMap(attrs map[string]interface{}, attributes []*common.KeyVa
...
-			attrs[attr.Key] = val
-			if truncatedBytes != 0 {
-				// if we trim a field, add telemetry about it; because we trim at 64K and
-				// a whole span can't be more than 100K, this can't happen more than once
-				// for a single span. If we ever change those limits, this will need to
-				// become additive.
-				attrs["meta.truncated_bytes"] = val
-				attrs["meta.truncated_field"] = attr.Key
-			}
+		if truncatedBytes := flatten(attrs, attr.Key, attr.Value, 0); truncatedBytes != 0 {
```
kentquirk, line 266: We believe that this limitation is no longer true in Shepherd (although it does need some testing to make certain that's the case). We should probably remove it -- could maybe be a different PR.

otlp/common.go
```diff
@@ -402,33 +402,56 @@ func getMarshallableValue(value *common.AnyValue) interface{} {
...
+			k := fmt.Sprintf("%s.%d", key, i)
+			if depth < maxDepth {
+				flatten(result, k, entry, depth+1)
+			} else {
+				truncatedBytes += addToMapAsJson(result, key, entry)
+			}
+		}
+	case *common.AnyValue_KvlistValue:
+		for _, entry := range value.GetKvlistValue().Values {
+			k := fmt.Sprintf("%s.%s", key, entry.Key)
```
kentquirk, line 429: In general, using Sprintf to concatenate things that are known to be strings is WAAAY more expensive than doing it with `+`, or for more complicated stuff, strings.Builder. It'd be worth changing this to just `+`.

Here's a benchmark:
```
BenchmarkSprintf-10          	  817777	      1474 ns/op	     108 B/op	       3 allocs/op
BenchmarkStringCat-10        	41451492	        29.89 ns/op	       0 B/op	       0 allocs/op
BenchmarkStringBuilder-10    	 2964910	       400.9 ns/op	      32 B/op	       2 allocs/op
```

```
var s1 = "prefix"
var s2 = "suffix"

func BenchmarkSprintf(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = fmt.Sprintf("%s.%s", s1, s2)
	}
}

func BenchmarkStringCat(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = s1 + "." + s2
	}
}

func BenchmarkStringBuilder(b *testing.B) {
	for i := 0; i < b.N; i++ {
		var sb strings.Builder
		sb.WriteString(s1)
		sb.WriteRune('.')
		sb.WriteString(s2)
		_ = sb.String()
	}
}
```

otlp/common.go
```diff
@@ -402,33 +402,56 @@ func getMarshallableValue(value *common.AnyValue) interface{} {
...
+// These types are all be marshalled to a string after conversion to Honeycomb-safe values.
+// We use our limitedWriter to ensure that the string can't be bigger than the allowable,
+// and it also minimizes allocations.
+// Note that an Encoder emits JSON with a trailing newline because it's intended for use
+// in streaming. This is correct but sometimes surprising and the tests need to expect it.
+func addToMapAsJson(attrs map[string]interface{}, key string, value *common.AnyValue) int {
+	val := getMarshallableValue(value)
+	w := newLimitedWriter(fieldSizeMax)
+	if err := json.NewEncoder(w).Encode(val); err != nil {
+		// TODOL: log error or report error when we have a way to do so
```
kentquirk, line 450:
```suggestion
		// TODO: log error or report error when we have a way to do so
```

otlp/common_test.go
```diff
@@ -52,48 +50,45 @@ func TestParseHttpHeadersIntoRequestInfo(t *testing.T) {
...
-			expected: nil,
+			expected: map[string]interface{}{},
 			attribute: &common.KeyValue{
 				Key: "", Value: &common.AnyValue{Value: &common.AnyValue_StringValue{StringValue: "str-value"}},
 			},
 		},
 		{
-			key:      "array-attr",
-			expected: "[\"one\",true,3]\n",
+			expected: map[string]interface{}{
```
kentquirk, line 87: I know this is an old test, but I've just realized OTel requires that arrays be homogenous types, so this is a test of something that can never happen.

otlp/common.go
```diff
@@ -402,33 +402,56 @@ func getMarshallableValue(value *common.AnyValue) interface{} {
...
-		enc := json.NewEncoder(w)
-		err := enc.Encode(arr)
-		if err == nil {
-			return w.String(), w.truncatedBytes
+		result[key] = value.GetIntValue()
+	case *common.AnyValue_BytesValue:
+		truncated := addToMapAsJson(result, key, value)
+		truncatedBytes += truncated
+	case *common.AnyValue_ArrayValue:
+		for i, entry := range value.GetArrayValue().Values {
```
kentquirk, line 419: Does it give us any qualms that this means that it becomes impossible to search an array for a value? I don't honestly know which is worse. Another option here is that we could add a DC function to let someone extract an indexed element of a JSON array.

otlp/common.go
```diff
@@ -43,6 +43,9 @@ const (
 // The limit is enforced in retriever (in private honeycomb code), in varstring.go.
 const fieldSizeMax = math.MaxUint16
 
+// max depth for flattening nested attributes
+const maxDepth = 10
```
kentquirk, line 47: This feels very deep to me. My off-the-cuff would have been 5.

otlp/logs_test.go
```diff
@@ -362,7 +394,15 @@ func TestCanExtractBody(t *testing.T) {
 					{Key: "key3", Value: &common.AnyValue{Value: &common.AnyValue_BoolValue{BoolValue: true}}},
 				},
 			}}},
-			expectedValue: "{\"key1\":\"value1\",\"key2\":2,\"key3\":true}\n",
+			expectedValue: map[string]interface{}{
+				"body.key1":        "value1",
+				"body.key2":        int64(2),
+				"body.key3":        true,
```
cartermp, line 400: For kvlists, I wonder if we should drop the `body` prefix.

Pros:

- Less prefixes which is nice maybe
- When using something like fluentbit, all attributes I add end up in the body even when not wanted (it's a bug they won't fix), and this would solve for that

Cons:

- Could have to de-dupe fields
- No way to distinguish between an attribute you added outside the log body and an attribute that lives within the log body (but does that even matter)

For arrays I think we would not do this though, since "0" and "1" are bad key names.

cartermp: Note: I don't have a strong preference either way, but it'll be a decision we take intentionally since it'll be breaking if we change back.

otlp/common.go
cartermp: We only support 5 levels for the events endpoint:
image

My take is -- let's match there. We can always go deeper if we need to, but it might be harder to back out if we need to.

otlp/common.go
cartermp: Another downside here is that we could potentially pollute a dataset with a thousand new columns when they never had any need for that data in the first place. KVLists are named, so that's at least more expected.

Hmmm. I'm more concerned with that downside than being able to search an array (as a string) for a value. Will have to let this simmer a little bit. We can probably proceed right now and then make a call after evaluating all the tradeoffs?

otlp/common.go
MikeGoldsmith: I know I knew this, but always find it hard to not try to avoid using `+` for string concatenation :|

MikeGoldsmith: Fixed in https://github.com/honeycombio/husky/pull/240/commits/70e188692fe32631aafc252e4f59bc86ac050fd4

otlp/common.go
MikeGoldsmith: Okay - I'll create a new ticket to investigate and we can remove in not required anymore.

MikeGoldsmith: Okay - I've created #241 to investigate if we still need to truncate.

otlp/common_test.go
MikeGoldsmith: Yeah, it shouldn't happen but also serves as a secondary test of different attribute values. No harm in keeping it 🤷🏻‍♂️

otlp/common.go
MikeGoldsmith: Updated to 5

otlp/logs_test.go
cartermp: I'll resolve this. Talked about it and it seems like there's consensus to keep the `body.` prefix.

kentquirk:
I like the options pattern, but instead of passing around the options as an extra parameter on all these function calls, what do you think of the idea of making these functions be methods on a type that carries the options? It's not significantly more efficient from a performance standpoint but it neatly separates the function signatures and also helps organize the code a bit.

MikeGoldsmith:
> I like the options pattern, but instead of passing around the options as an extra parameter on all these function calls, what do you think of the idea of making these functions be methods on a type that carries the options? It's not significantly more efficient from a performance standpoint but it neatly separates the function signatures and also helps organize the code a bit.

Interesting - so create a version of the translator with these options set. I'll try it out.

otlp/logs.go
```diff
@@ -56,13 +57,10 @@ func TranslateLogsRequest(request *collectorLogs.ExportLogsServiceRequest, ri Re
...
 				}
 				if log.Body != nil {
-					if val, truncatedBytes := getValue(log.Body); val != nil {
-						attrs["body"] = val
-						if truncatedBytes != 0 {
-							// if we trim the body, add telemetry about it
-							attrs["meta.truncated_bytes"] = truncatedBytes
-							attrs["meta.truncated_field"] = "body"
-						}
+					if truncatedBytes := flatten(attrs, options.bodyPrefix, log.Body, 0, options); truncatedBytes != 0 {
```
TylerHelmuth, line 60: This will result in new log bodies for customers.  Instead of the json string we'd create from the map, they'll get new fields on the event prefixed with `body`.  The end result, I believe, will be an event without a field named `body`.  That doesn't feel quite right to me.

We could leave the existing body conversion logic intact and do the new flatten logic in addition.  Would that be beneficial to customers? It would mean 1 less breaking change.

TylerHelmuth (commented):
In case it is useful, here is the flatten function we added recently for OTTL: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/ottlfuncs/func_flatten.go

otlp/logs.go
MikeGoldsmith: Correct, this currently replaces the single event field `body` with multiple event fields composed of the attributes found in the log body value.

If the log body attribute was a string, it would result in a single event field called `body`.

Yeah - I could add options to allow additive changes so we can enable/disable at our discretion 👍🏻

otlp/common.go
MikeGoldsmith: Decided to only do kv lists for now. We can revisit arrays later in #239.

kentquirk (approved):
Back to basics!  This seems correct to me.

TylerHelmuth (approved):
I like this simplified version

cartermp (approved):
Awesome work.

PR #245, opened 2024-02-29
feat: add Classic Ingest Key support

jharley:

We've now released Ingest Keys, but in order for them to work with Classic environments properly we need to update the key detection logic.

- Introduces a new public function `otlp.IsClassicApiKey()` to potentially be used in Refinery
- Renames the existing `(otlp.ReqInfo).hasLegacyKey()` method to `hasClassicKey` and updates it to use the above function

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206669928018720


otlp/common.go
```diff
@@ -48,7 +48,8 @@ const (
 const fieldSizeMax = math.MaxUint16
 
 var (
-	legacyApiKeyPattern = regexp.MustCompile("^[0-9a-f]{32}$")
+	legacyApiKeyPattern     = regexp.MustCompile("^[0-9a-f]*$")
+	classicIngestKeyPattern = regexp.MustCompile("^hc[a-z]ic_[0-9a-z]*$")
```
jharley, line 52: happy to adjust naming here

robbkidd: "Legacy" is a legacy term for "Classic" before the name "Classic" was settled on. I'm 👍🏻 with a whole-sale switchover to "classic" in code we update, especially as we add public functions/methods.

jharley: I'll update that public method name then 👍🏻

otlp/common.go
```diff
@@ -87,6 +88,16 @@ func IsContentTypeSupported(contentType string) bool {
 	return false
 }
 
+// IsClassicApiKey checks if the given API key is a Classic API key.
+func IsClassicApiKey(key string) bool {
+	if len(key) == 32 {
```
jharley, line 93: I remove the "len 0" check here as that wasn't being considered previously

MikeGoldsmith (approved):

PR #253, opened 2024-03-11
feat: Add tracing support

MikeGoldsmith:
Adds tracing support to Husky that can be configured by library users (eg beeline or otel). This is done by exposing a new global var `AddTelemetryAttributeFunc` that husky will call internally. The new func receives a context struct and a map of attributes with the library responsible to setting that telemetry on whatever telemetry client being used.

For example, a library user may configure the beeline then set the `AddTelemetryAttributeFunc` to add the attributes to the current beeline span like this:

```golang
husky.AddTelemetryAttributeFunc = func(ctx context.Context, key string, value any) {
	beeline.AddField(ctx, key, value)
}
```

I would like to make husky just use OTel for tracing, but some consumers are still using beeline and I don’t want to add beeline as a dependency. 

- Adds new `AddTelemetryAttributeFunc` that library users can set to receive telemetry fields about what husky is doing
- Updates all public funcs to also take a ctx object and pass along to internal funcs (eg `addAttributesToMap`)
- Set telemetry attributes when handling array, bytes or kvlist attributes
- Record the max depth of a kvlist as it's being traversed
- Add tests to verify telemetry attributes are called

### DIFF
--- a/telemetry.go
+++ b/telemetry.go
@@ -2,10 +2,15 @@ package husky
 
 import "context"
 
-var TracingFunc func(ctx context.Context, attributes map[string]any) = nil
+// SetAttributesFunc is a function that can be used to set attributes in telemetry controlled
+// by users of this package.
+// For example, beeline users would use beeline.AddField and OTel users would use span.SetAttributes.
+var SetAttributesFunc func(ctx context.Context, attributes map[string]any) = nil
 
-func AddAttributes(ctx context.Context, attributes map[string]any) {
-	if TracingFunc != nil {
-		TracingFunc(ctx, attributes)
+// SetAttributes is used internally to set attributes using the configured SetAttributesFunc.
+// This function is not intended to be used directly by consumers of this package.
+func SetAttributes(ctx context.Context, attributes map[string]any) {
+	if SetAttributesFunc != nil {
+		SetAttributesFunc(ctx, attributes)
 	}
 }
--- a/telemetry.go
+++ b/telemetry.go
@@ -1,0 +1,11 @@
+package husky
+
+import "context"
+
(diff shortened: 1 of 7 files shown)


telemetry.go
```diff
@@ -0,0 +1,16 @@
+package husky
+
+import "context"
+
+// SetAttributesFunc is a function that can be used to set attributes in telemetry controlled
```
JamieDanielson, line 5: Hmm naming is hard... maybe something like SetTelemetryAttributes?

JamieDanielson (commented):
This seems like a step in the right direction, though I would love to get 👀  from @kentquirk since this will be used in Refinery as well.

Also I'd recommend we add a note to the README asserting that this package is intended for internal Honeycomb use, or something of the sort, just to clearly distinguish it from other libraries that we expect and encourage folks to use.

telemetry.go
kentquirk: There's ambiguity: It's not "function that sets attributes", but "assign a function so that when I call SetAttributes it does the thing I want". So perhaps `DefineSetAttributesFunc`?

Or alternatively, rename `SetAttributes` to `AddAttributes` (since it's additive anyway) and call it `SetAddAttributesFunc`?

kentquirk, line 5:
```suggestion
// SetAttributesFunc is used to provide a function that adds attributes to telemetry controlled
```

kentquirk (approved):
I'm quite happy with this; I think we can find a better name, but it's definitely a win for Husky!

kentquirk (approved):
I think this is fine, given the architecture of Husky.

PR #259, opened 2024-05-08
feat(husky): Add in time since span event start from start of span

McSick:
It would be useful to know how long after the event happened from the span start e.g. opening and closing connection events. By adding this attribute, it becomes something a user could query.
-

Calculate time diff of span event time vs span start time.
-


MikeGoldsmith (approved):
I think this looks good.

PR #261, opened 2024-05-16
feat: Record span trace state as event field

MikeGoldsmith:
Records the span's tracestate value as an event field if present. Tracestate is use to share key-value pairs when propagating between services.

- Add span.TraceState value as event field if not empty
- Update unit test to verify field was added


kentquirk:
Given that tracestate is a set of k-v pairs, I feel like I'd like to see us parse it that way, and create a collection of values like "trace_state.k" with the value "v". 

As Robb notes in that thread, we should also think about whether this feature should be somehow controllable on a per-user basis. This is affecting data in the user's namespace.

TylerHelmuth (approved):
I am ok with ingesting the string value as-is (the key=value, comma delimited string) since that is the exact value.  Derived columns could take care of any parsing if necessary.

cartermp:
Hmmm, I'm more in favor of a "why not both" approach here:

```
trace.trace_state: ...
trace.trace_state.k1: v1
trace.trace_state.k2: v2
...etc.
```

The list of KVPs isn't likely to be terribly large, yes?

JamieDanielson:
> The list of KVPs isn't likely to be terribly large, yes?

That's an open question at the moment. It is easier to parse and add as extra fields later. It's much harder to take them away once they're added. Also they will potentially pollute database schema unexpectedly if someone doesn't want this.

PR #267, opened 2024-06-24
feat: 🌓🚀 add a utility attr name-to-description lookup function

robbkidd:

- Help folks know what an OTel attribute means.

The first baby steps toward showing a description for a recognized attribute. Look up names that match stable OpenTelemetry semantic conventions and return a reasonable description. The descriptions in this PR are pulled from the semconv "brief" associated with each attribute.

## The Future

- We expect to continue using the OTel semconv brief for each _stable_ attribute name.
- Generate this lookup from [the semconv models](https://github.com/open-telemetry/semantic-conventions/tree/main/model)?


MikeGoldsmith:
This has gone stale - closing for now.

PR #270, opened 2024-07-25
maint: appease codeql warning

codeboten:

The following warning is triggered by CodeQL scanning: "As of Go 1.21, toolchain versions must use the 1.N.P syntax." as per https://go.dev/doc/toolchain#version

- updated go.mod file to include the patch version as well. used .0 to remain compatible with all versions of go 1.21

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/honeycombio/husky
 
-go 1.21
+go 1.21.0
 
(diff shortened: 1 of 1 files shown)


JamieDanielson (approved):

JamieDanielson:
actually is there a later patch that we want here?

PR #275, opened 2024-09-19
feat: add sample rate to logs if set

### LINKED ISSUE #274: Set SampleRate for OTLP Logs
**Is your feature request related to a problem? Please describe.**

A [husky event](https://github.com/honeycombio/husky/blob/6a1cea3537c035344bb3e60b3001a5ec835b7cf8/otlp/common.go#L126-L130) is an object that contains both Attributes and Sample Rate (and Timestamp but that's irrelevant here).

```go
// Event represents a single Honeycomb event
type Event struct {
	Attributes map[string]interface{}
	Timestamp  time.Time
	SampleRate int32
}
```

[Traces add `SampleRate`](https://github.com/honeycombio/husky/blob/main/otlp/traces.go#L166) but logs do not.

**Describe the solution you'd like**

Set SampleRate for logs like we do for traces.

JamieDanielson:

- closes #274 

- move sample rate logic that was in `traces` to `common`, including test cases
- add `sampleRate` to log events
- make `sampleRate` key name case-insensitive
- fall back to default rate of 1 when given negative values 
- round and coerce floats into ints

It would be nice to have an error if an invalid sampleRateKey is found. Right now we just return empty string.

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -631,7 +631,7 @@ func getSampleRate(attrs map[string]interface{}) int32 {
 	}
 	// To make sampleRate consistent between Otel and Honeycomb, we coerce all 0 values to 1 here.
 	// Negative values are also invalid and so we convert to 1
-	if sampleRate == 0 || sampleRate < 0 {
+	if sampleRate <= 0 {
 		sampleRate = defaultSampleRate
 	}
 	delete(attrs, sampleRateKey) // remove attr
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -11,6 +11,7 @@ import (
 	"math"
 	"net/http"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
@@ -39,6 +40,8 @@ const (
 	gRPCAcceptEncodingHeader = "grpc-accept-encoding"
 	defaultServiceName       = "unknown_service"
 	unknownLogSource         = "unknown_log_source"
+
+	defaultSampleRate = int32(1)
 
 	// maxDepth is the maximum depth of a nested kvlist attribute that will be flattened.
 	// If the depth is exceeded, the attribute should be added as a JSON string instead.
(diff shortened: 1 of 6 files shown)


otlp/common.go
```diff
@@ -580,3 +583,55 @@ func shouldTrimTraceId(traceID []byte) bool {
 	}
 	return true
 }
+
+func GetSampleRate(attrs map[string]interface{}) int32 {
```
TylerHelmuth, line 587: This doesn't need to be exported, lets name it `getSampleRate`

otlp/common.go
```diff
@@ -580,3 +583,55 @@ func shouldTrimTraceId(traceID []byte) bool {
...
+	// A value of 1 means the span was not sampled
+	// For full explanation, see https://app.asana.com/0/365940753298424/1201973146987622/f
+	if sampleRate == 0 {
+		sampleRate = defaultSampleRate
+	}
+	delete(attrs, sampleRateKey) // remove attr
+	return sampleRate
+}
+
+func GetSampleRateKey(attrs map[string]interface{}) string {
```
TylerHelmuth, line 629: This doesn't need to be exported, lets name it `getSampleRateKey`

otlp/common.go
```diff
@@ -580,3 +583,55 @@ func shouldTrimTraceId(traceID []byte) bool {
...
+
+func GetSampleRate(attrs map[string]interface{}) int32 {
+	sampleRateKey := GetSampleRateKey(attrs)
+	if sampleRateKey == "" {
+		return defaultSampleRate
+	}
+
+	sampleRate := defaultSampleRate
+	sampleRateVal := attrs[sampleRateKey]
+	switch v := sampleRateVal.(type) {
```
kentquirk, line 595: what do you think about adding `float64` and `float32` to this, in case someone sends a value as '1.0`? This might happen if they're thinking about sampling probability instead of sample rate and have to do a division to get the value.

otlp/common.go
```diff
@@ -580,3 +583,55 @@ func shouldTrimTraceId(traceID []byte) bool {
...
+			if i < math.MaxInt32 {
+				sampleRate = int32(i)
+			} else {
+				sampleRate = math.MaxInt32
+			}
+		}
+	case int32:
+		sampleRate = v
+	case int:
+		if v < math.MaxInt32 {
```
kentquirk, line 607: if you're going to be this careful, you should probably guard against negative values too.

otlp/common.go
```diff
@@ -580,3 +583,55 @@ func shouldTrimTraceId(traceID []byte) bool {
...
+		}
+	case int64:
+		if v < math.MaxInt32 {
+			sampleRate = int32(v)
+		} else {
+			sampleRate = math.MaxInt32
+		}
+	}
+	// To make sampleRate consistent between Otel and Honeycomb, we coerce all 0 values to 1 here
+	// A value of 1 means the span was not sampled
```
kentquirk, line 620: "not sampled" means "dropped" -- I know this was copied code, but I would delete this line entirely. It doesn't add anything.

otlp/common.go
kentquirk, line 629: what do you think about expanding this to include all case variations by doing `strings.EqualFold` instead of this?

otlp/common_test.go
```diff
@@ -776,3 +778,68 @@ func Test_ReadOtlpBodyTooLarge(t *testing.T) {
...
+	})
+}
+
+func TestGetSampleRateConversions(t *testing.T) {
+	testCases := []struct {
+		sampleRate interface{}
+		expected   int32
+	}{
+		{sampleRate: nil, expected: 1},
+		{sampleRate: "0", expected: 1},
```
kentquirk, line 813: negative values?

kentquirk (approved):
I have some questions, but I trust however you want to handle them.

otlp/common.go
JamieDanielson: Great idea!

JamieDanielson:
> I have some questions, but I trust however you want to handle them.

I know you approved with suggestions but I wouldn't mind an extra check here. I made all the changes you suggested.

otlp/common.go
```diff
@@ -578,3 +581,68 @@ func shouldTrimTraceId(traceID []byte) bool {
...
+	case float64:
+		if v < math.MaxInt32 {
+			sampleRate = int32(v + 0.5)
+		} else {
+			sampleRate = math.MaxInt32
+		}
+	}
+	// To make sampleRate consistent between Otel and Honeycomb, we coerce all 0 values to 1 here.
+	// Negative values are also invalid and so we convert to 1
+	if sampleRate == 0 || sampleRate < 0 {
```
kentquirk, line 634: The compiler might do this anyway, but...
```suggestion
	if sampleRate <= 0 {
```

otlp/common.go
JamieDanielson: I added a check for floats also in the string conversion. I believe I made the change you suggested but take a look to confirm.

kentquirk (approved):
Looks great! One minor suggestion...

otlp/common.go
JamieDanielson: 🤦  Yep

PR #277, opened 2024-09-23
feat: Allow spans attributes to override service.name

MikeGoldsmith:

We currently only use the service.name value when it's set at the resource level and ignore it at the span attributes level. This change allows the service.name to be set via span attribute. The same service.name rules apply to both levels.

The precedence order is span attribute > resource attribute.

- https://github.com/honeycombio/pipeline-team/issues/163

- Allow service.name to be used for dataset from span attribute, overriding the resource level value if set
- Add unit tests

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -116,11 +116,9 @@ type TranslateOTLPRequestResult struct {
 }
 
 // Batch represents Honeycomb events grouped by their target dataset
-// SizeBytes is the total byte size of the OTLP structure that represents this batch
 type Batch struct {
-	Dataset   string
-	SizeBytes int
-	Events    []Event
+	Dataset string
+	Events  []Event
 }
 
 // Event represents a single Honeycomb event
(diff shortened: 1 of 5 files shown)


VinozzZ (changes requested):

JamieDanielson (commented):
My main concern with this is the breaking nature of the change. For many it's probably fine, but if anyone has a span with a `service.name` attribute that perhaps they've forgotten about or set in the past and forgotten to update, then this would result in their telemetry going to a different place than what is currently set in the resource attribute. 

As I understand it, the general idea of this change is intended to help folks who set a **span** attribute instead of a **resource** attribute to make sure their data is set to the proper service. Is it possible to do a check for `unknown_*` prefix before overwriting?

Theoretically this should only come up for new services being instrumented, after which they are fixed/updated to use resource attributes once the problem is discovered. By allowing the use of a span attribute, we're hiding the problem from the user and that could still have unexpected consequences outside of this (e.g. if they later use a collector processor that expects a `service.name` resource attribute they will have to fix it there anyway).

MikeGoldsmith:
I'm gonna close for now. We can revisit in the future if we still want to go this route.

PR #278, opened 2024-09-24
feat: add PHP to list of telemetry libraries

JamieDanielson:

- lack of visibility into php library usage

- add PHP instrumentation library prefix 
- fix typo in test for `telemetry.instrumentation_name`

### DIFF
--- a/otlp/traces_test.go
+++ b/otlp/traces_test.go
@@ -1107,12 +1107,12 @@ func TestInstrumentationLibrarySpansHaveAttributeAdded(t *testing.T) {
 	assert.Equal(t, "test_span_a", first_event.Attributes["name"])
 	assert.Equal(t, "First", first_event.Attributes["library.name"])
 	assert.Equal(t, "1.1.1", first_event.Attributes["library.version"])
-	assert.NotContains(t, first_event.Attributes, "telemtetry.instrumentation_library")
+	assert.NotContains(t, first_event.Attributes, libraryNameKey)
 	second_event := events[1]
 	assert.Equal(t, "test_span_b", second_event.Attributes["name"])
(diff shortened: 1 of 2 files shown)


otlp/traces_test.go
```diff
@@ -1107,12 +1107,12 @@ func TestInstrumentationLibrarySpansHaveAttributeAdded(t *testing.T) {
...
 	assert.Equal(t, "First", first_event.Attributes["library.name"])
 	assert.Equal(t, "1.1.1", first_event.Attributes["library.version"])
-	assert.NotContains(t, first_event.Attributes, "telemtetry.instrumentation_library")
+	assert.NotContains(t, first_event.Attributes, libraryNameKey)
 	second_event := events[1]
 	assert.Equal(t, "test_span_b", second_event.Attributes["name"])
 	assert.Equal(t, "io.opentelemetry.instrumentation.http", second_event.Attributes["library.name"])
 	assert.Equal(t, "2.2.2", second_event.Attributes["library.version"])
-	assert.Equal(t, true, second_event.Attributes["telemetry.instrumentation_library"])
+	assert.Equal(t, true, second_event.Attributes[libraryNameKey])
```
MikeGoldsmith, line 1115: I think we try to not use const's in testing to ensure the value is what we expect and it'll cause failures if the key is changed.

I'm not against using the const, but we have many instances where we don't use them so this one change makes it inconsistent.

MikeGoldsmith (approved):
Looks great. I would prefer not to use a const here as it wouldn't catch changes to the key like it would not.

otlp/traces_test.go
JamieDanielson: Gotcha. The only reason I had this change saved in a stash was because of the test on line 1110 that had a typo and so wasn't really testing the right thing. I'll pare this down to only change the typo.

PR #284, opened 2024-11-14
fix: revert back to recognize only "sampleRate" or "SampleRate" attributes

robbkidd:

1. A predicted performance regression during ingest: iterating over all attributes looking for a matching key name is much slower than 2 lookups into the attributes map.
2. Non-deterministic behavior in the probably-rare case of attributes containing multiple "samplerate" keys of differing capitalization, which would be astonishingly difficult to troubleshoot after ingest

- update the test to assert the previous behavior of recognizing only `sampleRate` and `SampleRate`
- … more coming

### DIFF
--- a/otlp/common_test.go
+++ b/otlp/common_test.go
@@ -787,18 +787,20 @@ func TestNoSampleRateKeyReturnOne(t *testing.T) {
 	assert.Equal(t, int32(1), sampleRate)
 }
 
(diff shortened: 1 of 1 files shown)


MikeGoldsmith (approved):

kentquirk (approved):
I love me a clear PR. Thanks.

PR #288, opened 2024-12-11
fix: Disable HTML escaping when JSON encoding values

MikeGoldsmith:

Husky JSON encodes attribute values in certain circumstances (eg deeply nested attributes). However, by default the JSON encoder escapes HTML characters which makes using the values more difficult later.

This change updates the JSON encoder to not escape HTML characters.

- Disable HTML escaping in the JSON encoder by calling `SetEscapeHTML(false)`
- Add unit test to verify an attribute that contains HTML is not escaped

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -460,9 +460,10 @@ func addAttributeToMap(ctx context.Context, result map[string]interface{}, key s
 // addAttributeToMapAsJson adds an attribute to a map as a JSON string.
 // Uses limitedWriter to ensure that the string can't be bigger than the maximum field size and
 // helps reduce allocation and copying.
-// Note that an Encoder emits JSON with a trailing newline because it's intended for use
+// Notes:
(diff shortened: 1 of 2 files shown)


otlp/common.go
```diff
@@ -462,10 +462,13 @@ func addAttributeToMap(ctx context.Context, result map[string]interface{}, key s
 // helps reduce allocation and copying.
 // Note that an Encoder emits JSON with a trailing newline because it's intended for use
 // in streaming. This is correct but sometimes surprising and the tests need to expect it.
+// NOTE: The JSON encoder has HTML escaping disabled.
```
codeboten, line 465: not a huge deal but this is the second note and the formatting doesn't match the previous `Note` comment

codeboten (approved):

TylerHelmuth (approved):

PR #297, opened 2025-04-29
feat: Allow OTLP/HTTP request translation to be provided max request body size

jparsons04:

When passing Husky Traces or Logs OTLP/HTTP request bodies, Husky currently defaults to a max request body size of 20MiB when the body is parsed. This can cause parsing issues if Husky is passed a file larger than 20MiB that contains batched trace or log telemetry within a single file. OTel collector batch processor configurations can reasonably exceed this limit for single files.

This PR creates some flexibility for the caller to define that limit while still [guarding against zipbombs](https://github.com/honeycombio/husky/pull/265) and maintaining existing functionality.

- Adds two new functions, `TranslateTraceRequestFromReaderSized` and `TranslateLogsRequestFromReaderSized` that allows a maxSize argument to be passed in
- Wraps `TranslateTraceRequestFromReader` and `TranslateLogsRequestFromReader` around these new functions, passing in the Husky-defined max request body size limit of 20 MiB

The end result should be a no-op. All current callers of `TranslateTraceRequestFromReader` and `TranslateLogsRequestFromReader` will continue to have their request bodies limited to 20 MiB. The new functions will be opt-in.

### DIFF
--- a/otlp/logs.go
+++ b/otlp/logs.go
@@ -13,11 +13,18 @@ import (
 // TranslateLogsRequestFromReader translates an OTLP log request into Honeycomb-friendly structure from a reader (eg HTTP body)
 // RequestInfo is the parsed information from the gRPC metadata
 func TranslateLogsRequestFromReader(ctx context.Context, body io.ReadCloser, ri RequestInfo) (*TranslateOTLPRequestResult, error) {
+	return TranslateLogsRequestFromReaderSized(ctx, body, ri, defaultMaxRequestBodySize)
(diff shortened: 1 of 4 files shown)


codeboten (approved):

kentquirk (approved):

jparsons04:
@codeboten @kentquirk FYI I've just added some unit tests that should cover the new functions.

TylerHelmuth (approved):

Issue #296, opened 2025-04-11
Support OTel standard for expressing sampling probability in trace state


kentquirk:
**Is your feature request related to a problem? Please describe.**

OTel has [standardized how sampling information should be transferred](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md). We should support it as it will soon start arriving in spans near you.

PR #301, opened 2025-05-12
feat: Add support for OTel sampling probability

MikeGoldsmith:

Add support for parsing OpenTelemetry's sampling probability that is part of the trace state.

- Closes https://github.com/honeycombio/husky/issues/296

- Add dependency on github.com/open-telemetry/opentelemetry-collector-contrib/pkg/sampling package to use the threshold parsing funcs
- Update getSampleRate to check for honeycomb attribute then fall back to otel trace state probability based sample rate
- Add tests


otlp/common.go
```diff
@@ -588,10 +590,28 @@ func shouldTrimTraceId(traceID []byte) bool {
 }
 
 // Sample Rate must be a whole positive integer
-func getSampleRate(attrs map[string]interface{}) int32 {
+func getSampleRate(attrs map[string]any, traceState string) int32 {
+	// Use Honeycomb's sampleRate attribute if it exists
+	sampleRate, ok := getSampleRateFromHoneycombAttribue(attrs)
```
kentquirk, line 595: Please fix the spelling of this function (missing `t`)

otlp/common.go
```diff
@@ -640,7 +660,37 @@ func getSampleRate(attrs map[string]interface{}) int32 {
...
+}
+
+// getSampleRateFromOTelSamplingThreshold returns the sampling threshold from the OpenTelemetry
+// trace state. Sampling threshold is a string that represents the sampling
+// probability using the OpenTelemetry sampling probability format.
+//
+// See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md
+func getSampleRateFromOTelSamplingThreshold(traceState string) (int32, bool) {
+	// split the trace state into key-value pairs
+	kvPairs := strings.Split(traceState, ",")
```
kentquirk, line 673: The pairs are separated by `;`, not `,` -- please fix.

otlp/common_test.go
```diff
@@ -870,3 +871,35 @@ func TestAddAttributesToMapAreNotHTMLEncoded(t *testing.T) {
...
+			assert.Equal(t, tt.expected, sampleRate)
+		})
+	}
+}
+
+func TestSampleRatePrefersHoneycombAttribute(t *testing.T) {
+	attrs := map[string]interface{}{
+		"sampleRate": 10,
+	}
+	sampleRate := getSampleRate(attrs, "th=c") // "th=c" is a sampling threshold of 1/4 or 25%
```
kentquirk, line 903: I think we need to test a few more examples of sampling thresholds. These are taken from [otep 235](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/trace/0235-sampling-threshold-in-trace-state.md).

th=4 -- equivalent to th=40000000000000, which is a 25% rejection threshold, corresponding to a 75% sampling probability.
th=c -- equivalent to th=c0000000000000, which is a rejection threshold of 75%, corresponding to a sampling probability of 25%.
th=08 -- equivalent to th=08000000000000, which is a rejection threshold of 3.125%, corresponding to a sampling probability of 96.875%.
th=0 -- equivalent to th=00000000000000, which is a 0% rejection threshold, which means Always Sample.

In particular, Honeycomb doesn't support any sample rates between 1 and 2, but you can express those in OTel as sampling probabilities between 50% and 100%.

What we should do to approximate that is randomly choose between 1 and 2 based on the actual threshold received. In other words, if the effective probability is 75%, then we should randomly choose either 1 or 2 with equal probability. For the 08 example above, we should choose 1 most of the time but 2 about 6.25% of the time. 

The easiest way to do that is to generate a random value in the range and compare it to the probability in the trace state.

```
// I don't know the actual function name
prob := t.SamplingProbability()
sampleRate := 1
if prob > 0.5 {
  r := 0.5 + (rand.Float64()/2)
  if r > prob {
    sampleRate := 2
 }
}
```

otlp/logs.go
```diff
@@ -84,7 +84,7 @@ func TranslateLogsRequest(ctx context.Context, request *collectorLogs.ExportLogs
 				}
 
 				// get sample rate after resource and scope attributes have been added
-				sampleRate := getSampleRate(attrs)
+				sampleRate := getSampleRate(attrs, "")
```
kentquirk, line 87: I'm not entirely sure if this is even applicable to logs. Is it?

MikeGoldsmith: Logs don't pass in their associated span's trace state so we're just passing in "". It's currently using the top level func, but could be made to just use the Honeycomb sample rate one.

otlp/common.go
```diff
@@ -640,7 +661,45 @@ func getSampleRate(attrs map[string]interface{}) int32 {
...
+	if !ok {
+		return defaultSampleRate, false
+	}
+	// get the sampling threshold
+	t, err := sampling.TValueToThreshold(th)
+	if err != nil {
+		return defaultSampleRate, false
+	}
+
+	// randomly sample between the celling and floor values of the adjusted count
```
kentquirk, line 694: Statistically, this works out to either an integer if the adjusted count is exact (powers of 2) or a half-integer if the adjusted count is not exact. 

It means that people trying to send, say, 80% of their data will see our average numbers at 75% instead, and they will think that something is wrong.

I'm about to push a friendly commit that I hope you'll agree with.

MikeGoldsmith: Yeah, please feel free to contribute an alternative solution.

MikeGoldsmith:
Thanks for the update @kentquirk  - I'm happy with your changes. Let me know if there's anything else you'd like to discuss before approving.

kentquirk (approved):
I think we're both happy now. I appreciate you getting this done!

PR #308, opened 2025-07-23
feat: support JSON OTLP in new fast path

ianwilkes:
JSON is still a thing even in 2025 - some SDKs ONLY support sending the JSON flavor of OTLP. We'll need to handle this efficiently.

This extends the work in #307 with a JSON version mirroring the proto version. It's not great to have, in effect, two copies of the same business logic, but I think the language gymnastics we'd have to do to consolidate it between both implementations in an efficient way would make the code truly impenetrable. We'll have to have a lot of faith in our testing - `TestUnmarshalTraceRequestDirect_Complete` now includes JSON.

I'm using fastjson because jsoniter doesn't actually support zero-allocation object iteration. Unfortunately both libraries appear to be abandonware, with no modern alternative on the horizon.

This is much slower than protobuf, but that's to be expected, and at least it's in the same ballpark.
```
BenchmarkUnmarshalTraceRequestDirectMsgp/Protobuf-12         	  156210	      6451 ns/op	    3377 B/op	       5 allocs/op
BenchmarkUnmarshalTraceRequestDirectMsgp/JSON-12             	   71346	     16525 ns/op	    3430 B/op	       7 allocs/op
```

### DIFF
--- a/otlp/traces_direct_json.go
+++ b/otlp/traces_direct_json.go
@@ -8,21 +8,33 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
-	"fmt"
 	"math"
 	"strconv"
 	"time"
 
 	"github.com/honeycombio/husky"
 
-	jsoniter "github.com/json-iterator/go"
+	"github.com/valyala/fastjson"
 	trace "go.opentelemetry.io/proto/otlp/trace/v1"
 )
 
-var jsonConfig = jsoniter.Config{
-	EscapeHTML:                    false,
-	ObjectFieldMustBeSimpleString: true,
-}.Froze()
+var parserPool fastjson.ParserPool
+
+// reusableParser is a wrapper to safely return parser to pool
+type reusableParser struct {
+	p *fastjson.Parser
+}
+
+func (rp *reusableParser) Release() {
+	if rp.p != nil {
+		parserPool.Put(rp.p)
+		rp.p = nil
+	}
+}
+
+func getParser() *reusableParser {
+	return &reusableParser{p: parserPool.Get()}
+}
 
 // bytesEqual compares a byte slice with a string without allocation
 func bytesEqual(b []byte, s string) bool {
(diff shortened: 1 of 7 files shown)


otlp/traces_direct_test.go
```diff
@@ -487,29 +492,6 @@ func TestUnmarshalTraceRequestDirect_Complete(t *testing.T) {
 		},
 	}
 
-	data := serializeTraceRequest(t, req)
```
ianwilkes, line 490: Note to reviewers, most of this code is the same, just moved; ignoring whitespace in the diff view should help.

otlp/traces_direct_json.go
```diff
@@ -0,0 +1,1115 @@
...
+							}
+						})
+					}
+				}
+			}
+		}
+	}
+
+	// Set event name
+	eventAttr.addString([]byte("name"), name)
```
VinozzZ, line 976: It would be nice to share the same logic for how we handle span event between JSON and protobuf implementation

otlp/traces_direct_json.go
```diff
@@ -0,0 +1,1115 @@
...
+	resourceAttrs,
+	scopeAttrs *msgpAttributes,
+	sampleRate int32,
+	isError bool,
+	batch *BatchMsgp,
+) error {
+	eventAttr := msgpAttributesPool.Get().(*msgpAttributes)
+
+	// Set trace info
+	if len(traceID) > 0 {
```
VinozzZ, line 1032: Are trace_id and trace.parent_id not required fields on span links and span events?

otlp/traces_direct_json.go
```diff
@@ -0,0 +1,1115 @@
...
+				"kvlist_max_depth":          depth,
+			})
+
+			if depth < maxDepth {
+				// Flatten the kvlist
+				if err = unmarshalKvlistValueFlattenJSON(ctx, v, key, attrs, depth+1); err != nil {
+					return
+				}
+			} else {
+				// Max depth exceeded, JSON encode the whole thing
```
VinozzZ, line 309: Can we add a test that exercise this code path?

otlp/traces_direct_json.go
```diff
@@ -0,0 +1,1115 @@
...
+			val, err = strconv.ParseInt(string(strBytes), 10, 64)
+			if err == nil {
+				result = val
+			}
+
+		case "doubleValue":
+			result = v.GetFloat64()
+
+		case "bytesValue":
+			// Bytes are base64 encoded in JSON
```
VinozzZ, line 386: Can you add a test case for "bytesValue" and "arrayValue"?

ianwilkes: These are covered in TestUnmarshalTraceRequestDirect_Complete, see bytes_attr and array_attr.

VinozzZ: I think the `bytes_attr` and `array_attr` in the [test data](https://github.com/honeycombio/husky/blob/a566cf3beca0a84ad9e3b5ce7c1ca2ce232fd46e/otlp/traces_direct_test.go#L111), doesn't cover this code because this code is only called when a byte value or array value is nested inside a array value or a key value list

otlp/traces_direct_json.go
```diff
@@ -0,0 +1,1115 @@
...
+				return
+			}
+
+			attrs.addFloat64(key, floatVal)
+
+		case "bytesValue":
+			// Bytes are base64 encoded in JSON
+			strBytes := v.GetStringBytes()
+			var b []byte
+			b, err = base64.StdEncoding.DecodeString(string(strBytes))
```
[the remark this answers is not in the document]
ianwilkes: Guess not, if we don't care about passing along invalid hex, which we probably don't.

otlp/traces_direct_json.go
ianwilkes: Trace id, yes. Parent id, I'm not so sure. If it were missing, it would still be usable in some ways.

otlp/traces_direct_json.go
ianwilkes: Good catch, I'll add JSON to `TestUnmarshalTraceRequestDirect_NestedMapAttributes` which covers this.

otlp/traces_direct_json.go
ianwilkes: Another good catch, I'll add these to TestUnmarshalTraceRequestDirect_NestedMapAttributes as well.

otlp/traces_direct_test.go
```diff
@@ -1370,29 +1579,6 @@ func TestUnmarshalTraceRequestDirectJSON_ErrorHandling(t *testing.T) {
 			}`,
 			errorMsg: "invalid syntax",
 		},
-		{
```
VinozzZ: Why do we remove this?

PR #314, opened 2025-09-10
feat: add prefix used by Collector telemetry scopes to recognized list

robbkidd:

We would like to be able to assess the provenance of telemetry sent to Honeycomb. We maintain an allowlist of strings that identify the open source OpenTelemetry instrumentation libraries. The allowlist lets us copy instrumentation_scope.name values from incoming telemetry, but only for the OTel instrumentation, not custom names given to custom tracers, meters, or loggers.

- relocate test: this logic is in common, not traces
- renamed test property to something other than the name of the function under test
- add the prefix that identifies telemetry coming from the OpenTelemetry Collector's instrumentation

I opted to stop the prefix at "-collector" and not "-collector-contrib" because the shorter is:

1. sufficient to identify the open source OTel stuff
2. who knows, maybe things other than "-contrib" get appended later, but it'll still be OTel's

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -71,14 +71,15 @@ var (
 
 	// List of per-lang instrumentation library prefixes
 	instrumentationLibraryPrefixes = []string{
-		"io.opentelemetry",                            // Java,
-		"opentelemetry.instrumentation",               // Python
-		"OpenTelemetry.Instrumentation",               // .NET
-		"OpenTelemetry::Instrumentation",              // Ruby
-		"go.opentelemetry.io/contrib/instrumentation", // Go
-		"@opentelemetry/instrumentation",              // JS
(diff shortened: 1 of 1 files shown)


VinozzZ (approved):

VinozzZ:
> We are interested in checking this list in places outside of this package. I'd like to make `isInstrumentationLibrary()` a public function in husky/common.
> 
>     1. Is that agreeable?
> 
>     2. Shall I bonus that into this PR or open a new PR?

Oops, I just saw your question after I have approved the PR. Maybe that's an indication for a separate PR?

robbkidd:
> Maybe that's an indication for a separate PR?

They _are_ separate concerns, so I have nothing against opening a new PR ... except laziness.

PR #318, opened 2025-09-29
fix(otlp): fix nil pointer in addAttributeToMap

asdvalenzuela:

- We were encountering a nil pointer exception when processing log data. The exception occurred on line 456 of `addAttributeToMap` when attempting to call methods on a `nil` `entry.value`. ([Sentry issue](https://honeycomb.sentry.io/issues/6633918091/?alert_rule_id=14802393&alert_timestamp=1758951911828&alert_type=email&environment=production&notification_uuid=fd589003-cd01-440b-aba1-0005408deee1&project=4505903518777344&referrer=alert_email))

- I just added a couple nil checks before we attempt further processing of the value.


kentquirk (commented):
Sending nil telemetry apparently never occurred to me. I feel like we should also add a test case for these two cases to Test_getValue.

PR #329, opened 2025-11-26
Span events sample rate persistence

matthewtapps:

- Manually setting a "SampleRate" attribute on a Span correctly sets the SampleRate for Honeycomb's query re-hydration on that Span.
- This is not currently the case for SpanEvents - instead the value is set to 1 (the default value).

- Adds SampleRate attribute unmarshalling to SpanEvents, exactly how it is performed for Spans already.

### DIFF
--- a/otlp/traces_direct.go
+++ b/otlp/traces_direct.go
@@ -1521,6 +1521,12 @@ func unmarshalSpanEvent(
 	eventAttr.addAttributes(scopeAttrs)
 	eventAttr.addAttributes(resourceAttrs)
 
+	// If the SpanEvent has a non-zero sampleRate
+	if eventAttr.sampleRate != 0 {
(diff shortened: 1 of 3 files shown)


robbkidd:
> I'm doing some **mad science with span processors** where I am filtering a bunch of individual low-utility spans and emitting a single span that has attributes summarising the content of all of the individual filtered spans

I would like to subscribe to your newsletter.

🤔 Interesting. OK. We're going into a holiday weekend in the 🇺🇸, so we'll have a ponder on this behavior change soon. I see and am intrigued by your case!

matthewtapps:
I am in the process of writing a blog post delving a little deeper into the Rollup Span Processor, I'll gladly share it once it's published!

Enjoy the holiday weekend :)

PR #354, opened 2026-03-18
feat: additionally translate event_name proto field into name

TylerHelmuth:

Updates Log translation to set `name` for the LogRecord's `event_name`. This is in addition to the existing translation of `event_name` to `event.name`. This solves 2 problems:

- Allows seemless `group by name` queries with datasets that contain both Spans and OTel Events
- Allows seemless transition in UI features, such as the trace waterfall span view, which has special logic looking for `name` from span event. 

Like for `event.name`, all attributes take precedence.

- update logs translation to also set `name` with `event_name`. 
- upate test.


MikeGoldsmith (approved):

TylerHelmuth:
Closing for now since we handled the naming via in-product updates instead.

PR #355, opened 2026-05-01
fix(otlp): handle hex-encoded trace/span IDs in direct JSON path

VinozzZ:
## Summary

The direct JSON path unconditionally base64-decoded `traceId`, `spanId`, and `parentSpanId`. The OTLP JSON spec says these fields should be lowercase hex strings, but proto3 JSON encoding specifies `bytes` fields as base64. Clients following the OTLP JSON spec (e.g. OTel JS HTTP/JSON exporter) send hex strings, which the direct path double-decoded: treating the 32-char hex string as base64 yields 24 bytes, which then get hex-encoded to a 48-char garbage value instead of the original 32-char ID.

## Changes

- Add `decodeIDBytes(s string)` that detects hex vs base64 and dispatches accordingly, replacing the 5 hardcoded `base64.StdEncoding.DecodeString` calls for trace/span IDs in span and link parsing
- Add `isHexString(s string)` backed by a 256-entry `[256]bool` lookup table; early-exits on `=` (base64 padding) before scanning
- Add `HexStringIDs` sub-test verifying that hex-encoded IDs round-trip correctly through the direct JSON path, matching proto path output

## Test plan

- `go test ./otlp/ -run TestUnmarshalTraceRequestDirect`
- New `HexStringIDs` sub-test covers the reported customer scenario (OTel JS sending hex trace/span IDs via OTLP HTTP/JSON)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/otlp/traces_direct_json.go
+++ b/otlp/traces_direct_json.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"encoding/hex"
 	"errors"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


otlp/traces_direct_json.go
```diff
@@ -1195,3 +1191,34 @@ func unmarshalSpanLinkJSON(
 	batch.Events = append(batch.Events, event)
 	return nil
 }
+
+// decodeIDBytes decodes a trace/span ID that may be hex (OTLP JSON spec) or base64 (proto3 JSON).
+func decodeIDBytes(s string) ([]byte, error) {
+	if isHexString(s) {
+		return hex.DecodeString(s)
+	}
+	return base64.StdEncoding.DecodeString(s)
```
TylerHelmuth, line 1200: Instead of needing `isHexString` could we default to hex string decoding and if the function returns an error, fall back to base64 decode?

VinozzZ: That's about 2-5x slower since decoding always allocates a new byte array. Since we are doing this for every single event, I think the performance matters

JamieDanielson (approved):

TylerHelmuth (approved):

PR #359, opened 2026-07-30
feat: composable header validation, support migrated environments

robbkidd:
## Problem

A classic API key normally requires a dataset header, since it's the only way husky knows which dataset to write events into. Some environments, though, get migrated in place to behave like Environments & Services (E&S). The customer keeps their classic key, but the environment now assigns datasets by `service.name`, the same way E&S already works. husky's validators don't know about that migration today, so a migrated environment gets rejected for doing exactly what an E&S key can already do. We need to split validation into what can happen before we know the Classicness of an environment (no longer just the key, but also the environment name) and what can only happen after auth (when we learn the environment name). 

Separately, the four existing validators (`ValidateTracesHeaders`, `ValidateMetricsHeaders`, `ValidateLogsHeaders`, `ValidateProfilesHeaders`) duplicate the same content-type and API-key checks, and two of them duplicate the dataset-header check verbatim.

### Composable validation

To support Classic environments migrated in place, we need to split the header validations into what needs to be run before authentication and what can only run after authentication.

- Add `ValidateHeaders()` for before auth, primarily to check API key, but also any other check common across signals (content-type)
- Add `ValidateDatasetHeader()` for after auth, because we don't know if the dataset header is required until we know the environment name which we get from auth
  - requires a dataset header if classic key is in use and the environment has not been migrated (determined with new `EnvironmentName` below)

- Add `RequestInfo.EnvironmentName`
  - set by callers after an environment name can be determined (generally after `ValidateHeaders` and auth)
  - blank means the environment behaves as Classic
  - non-blank on a classic key means the environment migrated in place

- Update `getDataset()`/`getDatasetFromMsgpAttr()`
  - follows the same "classic: dataset" vs "migrated: service.name" rule

### Deprecated signal-specific validators

I did _not_ refactor the existing validators to use the new functions. Kept the same logic, so a caller that upgrades husky without touching its call sites keeps current behavior: a classic key still requires the dataset header in those validators, migrated or not.

**I _did_ update several of the transform functions.** These called the signal-specific validators within them and are now updated to use the new composable functions. Callers that don't validate separately on their own before calling transform get the new migrated-environment behavior. Validating separately before calling transform is the new recommendation, though, and can be adopted once a project updates to a version of husky with this change.

### Tests

- Consolidate the duplicate `ValidateTracesHeaders`/`ValidateMetricsHeaders` tables to make clear "these are the same"
- Add coverage for:
  - the new functions across all `EnvironmentName` cases (classic and migrated, crossed with header present and absent, plus E&S)
  - the unchanged deprecated validators' behavior (ignores `EnvironmentName`)
  - the transform functions' new behavior on both the proto and msgpack/direct paths

## Non-breaking

`EnvironmentName`'s zero value reproduces current behavior, so this ships without _requiring_ any caller to change.

### DIFF
--- a/otlp/common.go
+++ b/otlp/common.go
@@ -170,6 +170,12 @@ type RequestInfo struct {
 	ApiKey  string
 	Dataset string
 
+	// EnvironmentName is the name of the environment associated with ApiKey,
+	// as reported by auth (e.g. the /1/auth endpoint). Callers set it after
+	// authenticating the key. Blank means the environment behaves as Classic;
+	// non-blank on a classic ApiKey means the environment migrated in place.
+	EnvironmentName string
+
 	UserAgent          string
 	ContentType        string
 	ContentEncoding    string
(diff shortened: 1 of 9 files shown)


otlp/common.go
```diff
@@ -209,28 +261,30 @@ func (ri *RequestInfo) ValidateMetricsHeaders() error {
 }
 
 // ValidateLogsHeaders validates required headers/metadata for a logs OTLP request
+//
+// Deprecated: use ValidateHeaders. Logs do not require a dataset header:
```
VinozzZ, line 265: I wonder if it's more explicit to say that `for log signal, it does not need to call `ValidateDatasetHeader` since logs do not require a dataset header`

VinozzZ:
I'm curious about the intention behind keeping the existing validator functions frozen. If we modified them directly — as this PR already does for the transform functions — users of husky could just bump the version and get the correct behavior, since the zero value of `EnvironmentName` preserves today's behavior for anyone not setting it.

With the current approach, we instead have to find every explicit validator call site and migrate it to the new API.

robbkidd:
> the intention behind keeping the existing validator functions frozen

@VinozzZ Ha haaaa, I managed to skip the need entirely in my explanation about deprecating the old validators. I'll try here and if it makes sense, I'll update the PR description.

As you describe, it's pretty trivial to update the existing validators to support the old behavior, when environment name is blank. I couldn't find a way to update them to support the new Fancy Classic way.

- To know whether a dataset header is required, we need to check the environment name.
- To know the environment name, we need to auth.
- To auth, we need the API key.
- To know we have an API key, we need to validate the headers.

So the options I saw were to:
1. split the header validations to functions that can be used before and after auth
2. push the act of determining the environment name into the validation functions somehow

№2 got very messy, so I didn't take it far before going with №1.

VinozzZ:
Thank you for the thorough explanation! That makes a lot sense to me now. Yea, the user of this library will have to make code changes to their header validation logic regardless. This change looks good to me!

VinozzZ (approved):

MikeGoldsmith (approved):

